PR #8, opened 2018-05-22
Récupération des amendements du Sénat

ronnix:


zam/aspirateur/zam_aspirateur/models.py
```diff
@@ -0,0 +1,51 @@
...
+        return int(mo.group(1))
+
+    def url(self, session, num_texte):
+        return "http://www.senat.fr/amendements/{}/{}/{}".format(
+            session,
+            num_texte,
+            self.html_page,
+        )
+
+    def evolve(self, **changes) -> 'Amendement':
```
davidbgk: Why not `    evolve = replace` if it's for the name?

zam/aspirateur/zam_aspirateur/fetcher.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Récupère les amendements à discuter, dans l'ordre de passage
+
+    NB : les amendements jugés irrecevables ne sont pas inclus.
+    """
+    assert phase in ('commission', 'seance')
+
+    url = f'http://www.senat.fr/en{phase}/{session}/{num}/liste_discussion.json'  # noqa
+
+    resp = requests.get(url)
```
davidbgk: I see a pattern here, not sure if it's worth extracting for 2 (maybe 3) cases.

zam/aspirateur/zam_aspirateur/fetcher.py
```diff
@@ -0,0 +1,75 @@
...
+    if resp.status_code == 404:
```
davidbgk: I used to prefer `http.HTTPStatus.NOT_FOUND` for explicitness.

zam/aspirateur/zam_aspirateur/fetcher.py
```diff
@@ -0,0 +1,75 @@
...
+        raise NotFound(url)
+
+    data = json.loads(resp.content)
```
davidbgk: Why not `resp.json()`?

zam/aspirateur/zam_aspirateur/fetcher.py
```diff
@@ -0,0 +1,75 @@
...
+
+    NB : les amendements jugés irrecevables ne sont pas inclus.
+    """
+    assert phase in ('commission', 'seance')
+
+    url = f'http://www.senat.fr/en{phase}/{session}/{num}/liste_discussion.json'  # noqa
+
+    resp = requests.get(url)
+    if resp.status_code == 404:
+        raise NotFound(url)
```
davidbgk: Another option is `resp.raise_for_status()`

zam/aspirateur/zam_aspirateur/amendements/writer.py
```diff
@@ -0,0 +1,87 @@
...
+    wb.save(filename)
+    return nb_rows
+
+
+def _write_header_row(ws: Worksheet) -> None:
+    header_row = [
+        FIELDS_NAMES.get(field, field.capitalize())
+        for field in FIELDS
+    ]
+    for column, value in enumerate(header_row, 1):
```
davidbgk: I learned that `enumerate` can take a second parameter today 😍

zam/aspirateur/zam_aspirateur/amendements/parser.py
```diff
@@ -0,0 +1,51 @@
+from datetime import datetime
+
+from ..clean import clean_html
+
+from .models import Amendement
+
+
+def parse_from_csv(d_amend: dict) -> Amendement:
```
davidbgk: I'm not sure it's parsing from csv anymore? Maybe `Amendement.from_csv` and `.from_json`?

zam/aspirateur/zam_aspirateur/amendements/writer.py
```diff
@@ -0,0 +1,87 @@
...
+            file_,
+            fieldnames=FIELDS,
+            delimiter=';',
+            quoting=csv.QUOTE_MINIMAL,
+        )
+        writer.writeheader()
+        for amendement in amendements:
+            writer.writerow(amendement.as_dict())
+            nb_rows += 1
+    return nb_rows
```
davidbgk: Isn't it the same as `len(amendements)`?

zam/aspirateur/zam_aspirateur/senateurs/parse.py
```diff
@@ -0,0 +1,25 @@
...
+            qualite=row['Qualité'],
+            nom=row['Nom usuel'],
+            prenom=row['Prénom usuel'],
+            groupe=row['Groupe politique'],
+        )
+        for row in csv.DictReader(lines)
+    ]
+    by_name = {}
+    for s in senateurs:
+        by_name[f"{s.qualite} {s.nom}".upper()] = s
```
davidbgk: Are we sure we never have homonyms?

zam/aspirateur/zam_aspirateur/__main__.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def process_amendements(
+    amendements: Iterable[Amendement],
+    session: str,
+    num: str,
+) -> Iterable[Amendement]:
+
+    # Les amendements discutés en séance, par ordre de passage
+    print(f'Récupération des amendements soumis à la discussion...')
```
davidbgk: Do we keep that pattern of beginning each string by `f`?

zam/aspirateur/setup.py
```diff
@@ -0,0 +1,13 @@
+from setuptools import setup
+
+
+setup(
+    name='zam-aspirateur',
```
davidbgk: Alright, I'll go that way with the client, I wasn't sure how we want to do this.

zam/aspirateur/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Pour récupérer la liste des amendements au PLFSS 2018 lors de la première lecture au Sénat (texte nº63) :
+
+```
+$ zam-aspirateur --session=2017-2018 --texte=63
+Wrote 595 rows to amendements_2017-2018_63.csv
+```
+
+Le résultat sera un fichier CSV comprenenant l'ensemble des amendements
+déposés, triés par ordre de passage lores de la discussion en séance.
```
davidbgk: lors*

zam/aspirateur/zam_aspirateur/models.py
ronnix: I was a bit sloppy when refactoring from `attrs` to `dataclasses`. I thought I would get back to it later. I guess later is now. 😄

zam/aspirateur/zam_aspirateur/fetcher.py
ronnix: Agree. I think I'll leave it like this for now.

zam/aspirateur/zam_aspirateur/fetcher.py
ronnix: Wow, new in 3.5, I had missed those. Neat!

https://docs.python.org/3/library/http.html#http.HTTPStatus

zam/aspirateur/zam_aspirateur/fetcher.py
ronnix: That would be simpler, indeed.

zam/aspirateur/zam_aspirateur/fetcher.py
ronnix: Yes, but I'd like the exception raised here to be a business one, not a low-level HTTP one.

But then maybe the `NotFound` name + the `url` param are confusing. Perhaps `LawNotFound` or `NoSuchLaw` would be more explicit?

zam/aspirateur/zam_aspirateur/amendements/parser.py
ronnix: Hmm, the `parse_from_csv` name is not great... Its not really parsing a CSV file, its more building an Amendement instance based on a CSV row. Maybe `build_from_csv_row` then?

zam/aspirateur/zam_aspirateur/amendements/writer.py
ronnix: `amendements` is an `Iterable[Amendement]` here, so it might be an iterator, not a sequence, in which case we can only consume it once

zam/aspirateur/zam_aspirateur/senateurs/parse.py
ronnix: I started with last name only as a key, but some senators do have the same name.

I found that in the CSV file, the author is given by last name only when it's not ambiguous (`M. FRASSA`), and by first name + last name when it would be ambiguous (`M. DANIEL LAURENT`). So I added that format as a key too.

I trust them not to be ambiguous, but that might be too optimistic. If they use an ambiguous name (`M. LAURENT`) we might pick up the wrong one.

It would probably be safer to extract the unique senator ID from the URL in the `"Fiche Sénateur"` field (it looks like `"//www.senat.fr/senfic/frassa_christophe_andre08018u.html" and `08018U` can be used as a unique key to find the senator.

ronnix: Fixed in 81e83a3d9ba67b1bee687b260cdbc125de6448bc

zam/aspirateur/zam_aspirateur/fetcher.py
ronnix: Fixed in a3b7029930104dcd1f4d37785f13322d78df7357

zam/aspirateur/zam_aspirateur/fetcher.py
ronnix: Fixed in b2da898d7ce421ba042eb4de7dc782fc4be7446f

zam/aspirateur/zam_aspirateur/models.py
ronnix: Fixed in afb31d24ab9400aba81b20eee97a394074a8a6c9

zam/aspirateur/zam_aspirateur/__main__.py
ronnix: I guess that was some bad copying and pasting. Fixed in d51291e8ae2e0b8c219da4458fc7df134c07fa1c.

zam/aspirateur/README.md
ronnix: Fixed in a2ad7398a898a200c7e652e8211adac43446bad7

PR #10, opened 2018-05-24
Connecting data from aspirateur to client viewer

davidbgk:
Still a WIP, we need to discuss pivotal data format and stable ids for amendements.


davidbgk:
A few things still scratching *me* on that PR:

* For the moment we connected formats mimicking the one from Drupal, I think it's suboptimal (to say the least) and we should try to find a pertinent modelisation for our case, especially handling both public/private data.
* Exchanging data from one Python script to another using a CSV/JSON file looks awkward, it might be more pertinent using for instance pickle or similar if we stay in the Python ecosystem or SQLite and [peewee](http://docs.peewee-orm.com/en/latest/index.html) or [dataset](http://dataset.readthedocs.io/en/latest/) on top of it if we want that result to be easily reusable.

Both are open discussions, we can go the current way for now and improve iteratively after but I had to write down what I have in mind to keep track of my reflections on that topic :).

See also: https://github.com/thombashi/sqlitebiter

ronnix:
I agree that we should move away from the Drupal output format, and design our own way to transfer information based on our own context and constraints.

With that said, if we just need one way communication from the aspirateur to the viewer, I like the idea of using JSON because it is easy to inspect, while SQLite is more opaque.

davidbgk:
Coincidence of the day:

> The Library of Congress Recommended Formats Statement (RFS) includes SQLite as a preferred format for datasets. — [Source](https://www.loc.gov/preservation/digital/formats/fdd/fdd000461.shtml)

ronnix:
Rebased the branch on top of master.

PR #11, opened 2018-05-25
Better styles/ergonomy based on users’ feedback

davidbgk:


zam/client/models.py
```diff
@@ -116,7 +117,8 @@ def load(
                     authors=authors,
                     group=group,
                     article=article,
-                    document=raw_amendement['document']
+                    document=raw_amendement['document'],
+                    is_gouvernemental=raw_amendement['gouvernemental']
```
ronnix: I like always adding a trailing comma, even on the last item, as it makes for nicer git diffs when adding or removing items, and prevents mistakes too

PR #16, opened 2018-05-31
Import Krill (WIP)

ronnix:


zam/webapp/zam_webapp/views/upload_krill.py
```diff
@@ -0,0 +1,160 @@
...
+
+        visionneuse_env = {
+            "ZAM_ASPIRATEUR_SOURCE": filename,
+            "ZAM_JAUNES_SOURCE": tmp_dir,
+        }
+        with environ(visionneuse_env):
+            _, aspirateur_items = load_aspirateur_source()
+            articles, amendements, reponses = load_data([], aspirateur_items)
+            html = render_and_save_html("Titre", articles, amendements, reponses)
+            response = Response(html)
```
davidbgk: I'm not sure it should be the responsibility of the repondeur to build that view, to me it should redirect to the visionneuse with a link to the generated json (not yet(?) implemented on the visionneuse side), thoughts?

PR #19, opened 2018-06-06
Load articles content from the output of tlfp

davidbgk:
You need to use the `parse-texte URL > tlfp-output.json` command from https://github.com/regardscitoyens/the-law-factory-parser/pull/77 to generate the output that can be loaded through the visionneuse.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,62 @@
+#!/bin/bash -e
+
+GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
+NC='\033[0m'
+
+message() {
+    echo
+    echo -e "${GREEN}>>> $1${NC}"
+}
+
+VENVS=/tmp/zamtest/virtualenvs
+mkdir -p "$VENVS"
+
+OUTPUT=/tmp/zamtest/output
+mkdir -p "$OUTPUT"
+
+message "Installation de l'aspirateur"
+[[ ! -d "$VENVS/aspirateur" ]] && python3 -m venv "$VENVS/aspirateur"
+pushd aspirateur > /dev/null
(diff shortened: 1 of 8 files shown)


test.sh
```diff
@@ -0,0 +1,62 @@
...
+popd > /dev/null
+
+message "Installation du parser de la fabrique de la loi"
+[[ ! -d "$VENVS/tlfp" ]] && python3 -m venv "$VENVS/tlfp"
+PIP="$VENVS/tlfp/bin/pip"
+$PIP install --quiet --upgrade pip
+$PIP install --quiet -e git+https://github.com/regardscitoyens/anpy.git#egg=anpy
+$PIP install --quiet -e git+https://github.com/regardscitoyens/legipy.git#egg=legipy
+$PIP install --quiet -e git+https://framagit.org/mdamien/metslesliens.git@1.1.1-manifest#egg=metslesliens
+$PIP install --quiet -e git+https://github.com/regardscitoyens/the-law-factory-parser.git#egg=tlfp
```
mdamien: You don't need to instead anpy/legipy/.. if you use the '--follow-links' option of pip. Soon, the flag won't be useful anymore since we're uploading everything to pypi but that's the temporary way to do it

davidbgk: I guess you mean `--process-dependency-links`? It works but displays a red message `DEPRECATION: Dependency Links processing has been deprecated and will be removed in a future release.`, that might be acceptable as a temporary workaround though.

Relevant threads:

* https://stackoverflow.com/questions/26061610/pip-install-dependency-links
* https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py

mdamien: Yes, that’s the good name, my bad 😅

PR #21, opened 2018-06-06
Add selectolax dependency to aspirateur setup.py

ronnix:


mdamien:
FYI, we had a similar problem and that's how we've done it for now: https://github.com/regardscitoyens/senapy

PR #24, opened 2018-06-12
Le répondeur récupère et affiche la liste des amendements, puis permet d'ajouter des réponses

ronnix:
capture d ecran 2018-06-12 a 13 10 38
capture d ecran 2018-06-12 a 13 10 44
capture d ecran 2018-06-12 a 13 10 48
capture d ecran 2018-06-12 a 13 11 08
capture d ecran 2018-06-12 a 13 11 13
capture d ecran 2018-06-12 a 13 11 47


repondeur/zam_repondeur/views/reponse.py
```diff
@@ -0,0 +1,53 @@
...
+            raise HTTPNotFound
+
+    @view_config(request_method="GET")
+    def get(self) -> dict:
+        return {"amendement": self.amendement, "avis": AVIS}
+
+    @view_config(request_method="POST")
+    def post(self) -> Response:
+        self.amendement.avis = self.request.POST["avis"]
+        self.amendement.observations = self.request.POST["observations"]
```
davidbgk: Just curious, don't you need to save explicitly your modifications at some point? Magic?

ronnix: Excellent question, sir.

There is actually some kind of "magic" happening, and you can see it in these two places :

- First, in `__init__.py` ([link](https://github.com/betagouv/zam/blob/repondeur-db/repondeur/zam_repondeur/__init__.py#L15)), when initializing the app, we add [pyramid_tm](https://docs.pylonsproject.org/projects/pyramid-tm/en/latest/), which is an extension connecting the Pyramid request/response cycle with the [transaction](https://pypi.org/project/transaction/) package, which is itself a generic *transaction manager*.

- Second, in `models/base.py` ([link](https://github.com/betagouv/zam/blob/repondeur-db/repondeur/zam_repondeur/models/base.py#L6)), we also add an extension to the SQLAlchemy session factory, so the it is attached to the transaction framework.

The way it works then, is that a new transaction is started for each request, and if that request succeeds, then the transaction will be committed, and the changes will be saved to the DB ; if the request fails, though (i.e. an exception is raised), the transaction will be rolled back, and the changes will be canceled.

This is different from Django where you would do an explicit `save()`, or from the way SQLAlchemy is typically used with Flask, where you would do a explicit `db_session.commit()`.

This pattern for transaction management is indeed one of the reasons I like Pyramid a lot.

admin/fabfile.py
```diff
@@ -65,32 +66,146 @@ def bootstrap(ctx):
...
+    install_requirements(ctx, app_dir=app_dir, user=user)
+    setup_config(ctx, app_dir=app_dir, user=user, secret=secret)
+    initialize_db(ctx, app_dir=app_dir, user=user)
+    fetch_an_group_data(ctx, user=user)
+    setup_service(ctx, an_pattern_liste, an_pattern_amendement)
+
+
+@task
+def create_user(ctx, name, home_dir):
+    if ctx.sudo(f"getent passwd {name}", warn=True, hide=True).failed:
```
davidbgk: Maybe having a dedicated `bootstrap_repondeur` command would avoid most of extra checks?

aspirateur/tests/textes/test_get_dossiers_legislatifs.py
```diff
@@ -0,0 +1,96 @@
...
+
+import pytest
+
+
+HERE = Path(os.path.dirname(__file__))
+
+
+@pytest.yield_fixture
+def sample_file():
+    path = HERE / ".." / "sample_data" / "Dossiers_Legislatifs_XV.json"
```
davidbgk: `HERE.parent`?

aspirateur/tests/textes/test_get_dossiers_legislatifs.py
```diff
@@ -0,0 +1,96 @@
...
+    with open(path) as f_:
```
davidbgk: `yield path.open()` should be enough.

admin/fabfile.py
```diff
@@ -65,32 +66,146 @@ def bootstrap(ctx):
...
+@task
+def setup_config(ctx, app_dir, user, secret):
+    template_local_file(
+        "../repondeur/production.ini.template",
+        "../repondeur/production.ini",
+        {"secret": secret},
+    )
+    sudo_put(
+        ctx, "../repondeur/production.ini", f"{app_dir}/production.ini", chown=user
+    )
```
davidbgk: Then delete the local `production.ini` generated file?

repondeur/README.md
```diff
@@ -11,6 +11,30 @@ $ pipenv install
...
+Get the groups data for Assemblée nationale:
+
+```
+$ curl --silent --show-error http://data.assemblee-nationale.fr/static/openData/repository/15/amo/deputes_actifs_mandats_actifs_organes_divises/AMO40_deputes_actifs_mandats_actifs_organes_divises_XV.json.zip -o groups.zip
+$ mkdir -p data/an/groups
+$ unzip -q -o groups.zip -d data/an/groups/
+$ rm groups.zip
+```
+
+Define the folowing environment variables:
```
davidbgk: `folLowing`

repondeur/tests/test_lectures_list.py
```diff
@@ -13,23 +8,11 @@ def test_get_list_empty(app):
...
-
 def test_get_list_not_empty(app, dummy_lecture):
 
     resp = app.get("/lectures/")
 
     assert resp.status_code == 200
     assert resp.content_type == "text/html"
 
-    assert "Sénat, session 2017-2018, texte nº 63" in resp.text
+    assert "Assemblée nationale, session 15, texte nº 0269" in resp.text
```
davidbgk: Makes me wonder if we want to left strip zeros on text numbers.

repondeur/zam_repondeur/templates/lectures_add.html
```diff
@@ -12,56 +18,73 @@ <h2>Ajouter une lecture...</h2>
...
-    </table>
-    {% endif %}
+{% endblock %}
+
+
+{% block scripts %}
+<script src="{{ request.static_url('zam_repondeur:static/choices/choices.min.js') }}"></script>
+<script type="text/javascript">
+
+    var lectures_by_dossier = {{ lectures_by_dossier | tojson | safe }};
```
davidbgk: Not sure where we want to go here with ES6/semicolumn-less/etc.

repondeur/zam_repondeur/templates/reponse_edit.html
```diff
@@ -0,0 +1,120 @@
...
+        }
+    </style>
+    <script src="{{ request.static_url('zam_repondeur:static/tinymce/tinymce.min.js') }}"></script>
+    <script type="text/javascript">
+        tinymce.init({
+            selector: "textarea",
+            language: "fr_FR",
+            menubar: false,
+            plugins: "lists fullscreen",
+            toolbar: "undo redo | bold italic underline strikethrough removeformat | bullist numlist | cut copy paste | fullscreen",
```
davidbgk: I would let `bold`, `italic` and `fullscreen` only for now. And then add more options only if real blockers. When I see how current answers are full of bold I'm afraid we will have meaningless styles otherwise…

repondeur/zam_repondeur/views/lectures.py
```diff
@@ -1,94 +1,65 @@
...
-        lecture_dir = os.path.join(self.data_dir, f"{chambre}-{session}-{num_texte}")
-        if os.path.isdir(lecture_dir):
+        dossier_uid = self.request.POST["dossier"]
+        texte_uid = self.request.POST["lecture"]
+
+        dossier = self.dossiers_by_uid[dossier_uid]
+        lecture = dossier.lectures[texte_uid]
+
+        if lecture.chambre == Chambre.AN:
+            chambre = "an"
```
davidbgk: I don't get why you don't reuse directly `lecture.chambre`?

repondeur/zam_repondeur/views/lectures.py
```diff
@@ -1,94 +1,65 @@
...
+            session = "15"
```
davidbgk: Couldn't we get that from the `chambre`?

admin/fabfile.py
ronnix: I understand that we may want to split between "things we only need to do once", and "things we need to do every time we deploy" for efficiency.

But then one day, to deploy the new version of the app, you need to add some new Debian package that the app now depends on. So you add it to the `bootstrap` task. But then to deploy that new version, you cannot just run `fab deploy`. You need to do `fab bootstrap` again, or it will fail.

So every time you deploy a new version, you need to think: what steps are needed this time? Do I need to run `bootstrap` again, or would `deploy` be enough?

Or you just run both `bootstrap` and `deploy` every time, just to be sure.

I don't like to think, so I kinda like having a "single command" idempotent deploy, to keep it simple and minimize errors, even if it's somewhat less efficient.

aspirateur/tests/textes/test_get_dossiers_legislatifs.py
ronnix: Fixed in f1cd03424abf0cd13f2f0434abdb09f38c24bb30

aspirateur/tests/textes/test_get_dossiers_legislatifs.py
ronnix: I wanted to make sure the file was properly closed after use by using a context manager here, but I'm just mocking the opening of the file here, and it will be closed in the `get_dossiers_legislatifs` function, so you're right we can make it simpler, and just a `return path.open()` will even do.

Fixed in 09c6821bc2fba4991a3cdfa79b81b41223aa7c32

admin/fabfile.py
ronnix: Good idea! I turned `template_local_file` into a context manager that will automatically remove the generated file after use → c7052df9609a613b43f73977a57f5a89c02a405b

repondeur/README.md
ronnix: Fixed in 9d5c65d8442b943a3c09d432d52a6645a99ac6f2

repondeur/tests/test_lectures_list.py
ronnix: I think those pseudo-titles will go away soon, and we'll just use the real title from the *projet de loi* instead.

repondeur/zam_repondeur/templates/lectures_add.html
ronnix: I'm not sure what the best practices are today. Do all important browsers support ES6 now?

(I think I do like semicolons though, must be all the time I spent writing C and C++?)

repondeur/zam_repondeur/templates/reponse_edit.html
ronnix: I'm removing `underline` and `strikethrough` → 4e250dd9d0dd661dfe37ae08e2fac5028b28fe6f

OK to keep the `bullist` and `numlist`?

repondeur/zam_repondeur/views/lectures.py
ronnix: You're right. Changed in 034a583d9e03f9a0bb22bc289a85edc3cf5cbba6.

repondeur/zam_repondeur/views/lectures.py
ronnix: Not sure what you have in mind here.

repondeur/tests/test_lectures_list.py
ronnix: I did some clean up anyway, and made sure all the texte and amendement numbers are integers now → 88f3bda044f6340a2262764e7de716f896ee374f

PR #32, opened 2018-06-22
Deployment

ronnix:
- Add option to wipe the DB when deploying the répondeur (the current DB will be backed up to `/var/backups/zam/` first)
- Pin the pipenv version on the server to avoid possible inconsistent behaviours

### DIFF
--- a/admin/fabfile.py
+++ b/admin/fabfile.py
@@ -104,7 +104,7 @@ def deploy(ctx, source):
 
 @task
 def deploy_repondeur(
-    ctx, secret, an_pattern_liste, an_pattern_amendement, branch="master"
+    ctx, secret, an_pattern_liste, an_pattern_amendement, branch="master", wipe=False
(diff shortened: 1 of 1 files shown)


admin/fabfile.py
```diff
@@ -160,6 +162,16 @@ def setup_config(ctx, app_dir, user, secret):
         )
 
 
+@task
+def wipe_db(ctx, user):
+    create_directory(ctx, "/var/backups/zam", owner=user)
+    ctx.sudo(
+        f"cp --backup=numbered /var/lib/zam/repondeur.db /var/backups/zam/repondeur.db",
```
davidbgk: Is that `--backup` option standard? (I don't have it on my machine)

ronnix: Standard in GNU `ls`, but not in the BSD-flavored one you get on macOS...

PR #39, opened 2018-06-25
Improve CSV & Excel exports 

ronnix:
- Convert HTML fields to text when exporting to CSV or Excel
- Use consistent vocabulary in CSV/Excel exports and edit form

### DIFF
--- a/aspirateur/zam_aspirateur/amendements/writer.py
+++ b/aspirateur/zam_aspirateur/amendements/writer.py
@@ -12,8 +12,6 @@ from openpyxl.worksheet import Worksheet
 from .models import Amendement
 
 
-FIELDS = [field.name for field in fields(Amendement)]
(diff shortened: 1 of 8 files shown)


aspirateur/zam_aspirateur/amendements/writer.py
```diff
@@ -78,7 +86,7 @@ def write_xlsx(amendements: Iterable[Amendement], filename: str) -> int:
 
 
 def _write_header_row(ws: Worksheet) -> None:
-    header_row = [FIELDS_NAMES.get(field, field.capitalize()) for field in FIELDS]
+    header_row = [rename_field(field_name) for field_name in FIELDS]
```
ronnix: Ici j'aurais dû mettre `HEADERS`

PR #47, opened 2018-06-26
(Des)integrate visionneuse into repondeur

davidbgk:


repondeur/zam_repondeur/models/visionneuse.py
```diff
@@ -0,0 +1,70 @@
...
+@dataclass
+class Article:
+    pk: str
+    id: int
+    titre: str
+    etat: Optional[str] = ""
+    multiplicatif: Optional[str] = ""
+    jaune: Optional[str] = ""
+    alineas: Optional[dict] = None
+    amendements: Any = field(default_factory=lambda: [])  # List[Amendement]
```
ronnix: Why the convoluted type annotation here?

davidbgk: Because you cannot reference `Amendement` which is declared below.

ronnix: Try putting in in double quotes, mypy should handle that.

davidbgk:
`−1,121` even better \o/

PR #49, opened 2018-06-26
Improve fuzzy search in lecture add form by using selectize

ronnix:
screencast

I wanted to avoid the jQuery dependency initially by using something more lightweight, but Selectize has a better fuzzy search implementation:

https://selectize.github.io/selectize.js/

Last release is from 2016, so it's either stable or unmaintained :)

If we ever need to change to something else, the other popular options are:
- Select2 (https://select2.org)
- Chosen (https://harvesthq.github.io/chosen/)


davidbgk:
The trunk is still active https://github.com/selectize/selectize.js/commits/master

repondeur/zam_repondeur/templates/lectures_add.html
```diff
@@ -18,19 +19,15 @@ <h2>Ajouter une lecture...</h2>
...
     <form action="{{ request.path }}" method="POST">
         <div class="form-group">
-            <label for="dossier">Texte de loi</label>
-            <select id="select-dossier" name="dossier" class="form-control">
-                {% for dossier in dossiers %}
-                    <option value="{{ dossier.uid }}" {%- if loop.first %} selected{% endif %}>{{ dossier.titre }}</option>
-                {% endfor %}
+            <label for="dossier">Dossier législatif</label>
+            <select id="select-dossier" name="dossier" class="form-control" placeholder="Choisir un dossier législatif...">
+                <option value="" selected="selected"></option>
```
davidbgk: Isn't it possible to keep the whole list here and flush it through JS on selectize load? It would make it accessible even if JS crashes somehow.

repondeur/zam_repondeur/templates/lectures_add.html
```diff
@@ -43,48 +40,56 @@ <h2>Ajouter une lecture...</h2>
...
+                    return;
+                }
+                $.ajax({
+                    url: "{{ request.route_url('choices_lectures', uid='UID') }}".replace("UID", uid),
+                    type: "GET",
+                    error: function() {
+                        alert("Impossible de récupérer la liste des lectures.\nMerci de réessayer plus tard !");
+                    },
+                    success: function(res) {
+                        for (let uid in res.lectures) {
```
davidbgk: I think that syntax is not consistent with the fact of using `var` elsewhere. We should discuss as a team which browsers/versions we want to support.

repondeur/zam_repondeur/views/lectures.py
```diff
@@ -324,3 +311,40 @@ def fetch_articles(request: Request) -> Response:
             num_texte=lecture.num_texte,
         )
     )
+
+
+@view_config(route_name="choices_dossiers", renderer="json")
+def choices_dossiers(request: Request) -> dict:
```
davidbgk: No tests?

repondeur/zam_repondeur/views/lectures.py
```diff
@@ -324,3 +311,40 @@ def fetch_articles(request: Request) -> Response:
...
+    return {
+        "dossiers": [
+            {"uid": uid, "titre": dossier.titre}
+            for uid, dossier in dossiers_by_uid.items()
+        ]
+    }
+
+
+@view_config(route_name="choices_lectures", renderer="json")
+def choices_lectures(request: Request) -> dict:
```
davidbgk: Same here.

ronnix:
Thanks for the feedback!

PR #53, opened 2018-06-28
Fix some typing issues in aspirateur

ronnix:


ronnix:
Merging is the way to go!

ronnix:
Also, I started looking into setting up Circle CI

PR #54, opened 2018-06-28
Handle pasting junk from Word

ronnix:


ronnix:
Good point, I had forgotten that one!

PR #62, opened 2018-07-01
Deal with datetimes for lectures and display a warning on update at visionneuse

davidbgk:
capture d ecran 2018-06-30 a 23 48 21

PR #66, opened 2018-07-04
Add fetching of amendements for commissions (Assemblée nationale)

ronnix:
This requires multiple things:

- being able to generate the appropriate AN URLs using the right short code (different for séance publique and each commission)
- being able to distinguish lectures and amendements by different commissions and/or séance publique on the same "texte" (i.e. adding `organe` to the primary key)


repondeur/tests/testapp.py
```diff
@@ -0,0 +1,21 @@
...
+from webtest import (
+    TestApp as BaseTestApp,
+    TestRequest as BaseTestRequest,
+    TestResponse as BaseTestResponse,
+)
+
+
+class TestResponse(BaseTestResponse):
+    @reify
+    def parser(self):
```
davidbgk: 👍

repondeur/tests/test_articles_add.py
```diff
@@ -139,25 +143,28 @@ def test_post_form_senat(app, dummy_lecture, dummy_amendements):
...
+
+        amendement = dummy_amendements[0]
         amendement.chambre = "senat"
         amendement.session = "2017-2018"
         amendement.num_texte = 63
+        amendement.organe = "PO78718"
         amendement.subdiv_num = "4"
         amendement.subdiv_mult = "bis"
 
+        # The objects are no longer bound to a session here, as they were created in a
```
davidbgk: 👌

repondeur/zam_repondeur/models/lecture.py
```diff
@@ -19,22 +22,52 @@ class Lecture(Base):  # type: ignore
...
+
+    def format_organe(self) -> str:
+        result: str = self.organe
+        organes = get_data("organes")
+        if self.organe in organes:
+            organe_data = organes[self.organe]
+            result = organe_data["libelleAbrege"]
+        return self.rewrite_organe(result)
+
+    def rewrite_organe(self, label: str) -> str:
```
davidbgk: I'm not sure if that's pertinent but there is that [particular line](https://github.com/betagouv/zam/blob/master/repondeur/zam_repondeur/fetch.py#L53) that might reuse it?

ronnix:
This PR is getting a bit large, so I’m willing to extract some bits into separate PRs if you’d like.

ronnix:
@davidbgk I removed the title macros, which makes for more consistent titles 👍

repondeur/zam_repondeur/models/lecture.py
ronnix: Hmm, I think the logic in that line might be wrong.

If the case of the Assemblée nationale, the text we want to get the articles of might be:
- the text as initially submitted → `f"{BASE_URL}{lecture.session}/projets/pl{lecture.num_texte:04}.asp"`
- the text as modified by a commission → `f"{BASE_URL}leg/pjl{lecture.session[2:4]}-{lecture.num_texte:03}.html"`

The URL pattern does not depend on whether the lecture is commission or séance publique, it depends on what happened before, and we don't keep that information handy... 😞

ronnix: Maybe the easy way is to just try both URLs and see which one works...

ronnix: Ok, so I did a separate PR for this (#67), tell me what you think.

davidbgk:
> This PR is getting a bit large, so I’m willing to extract some bits into separate PRs if you’d like.

It's alright, let's get it merged on Monday to wipe the DB only once on `zam` for both that PR and #62

PR #69, opened 2018-07-09
Do not display abandonned amendements

davidbgk:


repondeur/tests/test_reponses.py
```diff
@@ -46,6 +48,51 @@ def test_reponses_gouvernemental(app, dummy_lecture, dummy_amendements):
...
+        # Only the last one.
+        amendement.sort = sort
+        DBSession.add_all(dummy_amendements)
+
+    resp = app.get("http://localhost/lectures/an/15/269/PO717460/reponses")
+
+    assert resp.parser.tags("h1")[0].text() == str(dummy_lecture)
+    assert len(resp.parser.tags("section")) == 1
+    assert len(resp.parser.tags("article")) == 1
+    assert len(resp.parser.css("article.gouvernemental")) == 0
```
ronnix: It’s not obvious to me what we’re testing with these assertions

repondeur/tests/testapp.py
```diff
@@ -12,6 +12,13 @@ class TestResponse(BaseTestResponse):
     def parser(self):
         return HTMLParser(self.text)
 
+    @property
+    def title(self) -> str:
+        return self.parser.css_first("h1").text()
```
ronnix: Nice! When reading the tests I thought it was the contents of the <title> tag. Should we call it first_h1 or something to avoid the ambiguity?

repondeur/tests/testapp.py
```diff
@@ -12,6 +12,13 @@ class TestResponse(BaseTestResponse):
...
+
+    def __contains__(self, amendement) -> bool:
+        return bool(self.parser.css_first(f"#amendement-{amendement.num:06}"))
```
ronnix: Love the idea too. However when reading the test I was not sure what was happening exactly. Was the amendement being stringified maybe?

And what if we want to test the presence of other kinds of things on other pages? Maybe we could use a more explicit method name like contains_amendement?

ronnix:
Rebased

repondeur/tests/testapp.py
```diff
@@ -12,6 +12,23 @@ class TestResponse(BaseTestResponse):
...
+            return False
+        assert (
+            str(amendement.num) in anchor.parent.css_first("header h2").text().strip()
+        )
+        if is_gouvernemental:
+            assert anchor.parent.attributes == {"class": "gouvernemental"}
+        else:
+            assert anchor.parent.attributes == {}
+        return True
+
```
ronnix: I found the asserts in the middle of a boolean method, itself used within an assert, a bit confusing.

PR #75, opened 2018-07-10
Show deployed version (branch, sha1 & date) in HTML comment

ronnix:


davidbgk:
If that's only for us, why not put it as a HTML comment?

ronnix:
That was my initial idea, but then sometimes it's a pain to access the page source (e.g. on a tablet) so I thought I would make it easier to access. What do you think?

davidbgk:
Or maybe a dedicated page? I dislike when a developer info adds some bloat to the user

ronnix:
Ok, I'll make it hidden then!

PR #78, opened 2018-07-11
Merge aspirateur into repondeur

ronnix:
In a second step, we can do more refactoring, but this puts everything into the same project now.

Note: we had 2 different HTML cleaning functions, each allowing a different set of tags. I unified them, but this could cause some changes in what we display...


repondeur/zam_repondeur/clean.py
```diff
@@ -1,9 +1,15 @@
+from html import unescape
+from typing import Iterable, Optional
+
 import bleach
 
 
-ALLOWED_TAGS = ["p", "ul", "li", "b", "i", "strong", "em"]
+ALLOWED_TAGS = ["div", "p", "ul", "li", "b", "i", "strong", "em", "sup"]
```
davidbgk: Allow `sub` and `ol` for consistency?

aspirateur/setup.py
```diff
@@ -8,16 +8,5 @@
...
-        "cachecontrol",
-        "dataclasses",
-        "inscriptis",
-        "lockfile",
-        "openpyxl",
-        "requests",
-        "selectolax",
-        "xmltodict",
-    ],
+    install_requires=[],
```
davidbgk: Is the setup still required?

ronnix:
I was planning to do a `pipenv install` at the end to make sure everything is ok, but forgot. Maybe it was not required!

PR #80, opened 2018-07-12
Use Pyramid's resource tree traversal for routing

ronnix:
Reference: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/muchadoabouttraversal.html

### DIFF
--- a/repondeur/zam_repondeur/fetch/models.py
+++ b/repondeur/zam_repondeur/fetch/models.py
@@ -1,6 +1,6 @@
 import re
 from dataclasses import asdict, dataclass, fields, replace
-from datetime import date
+from datetime import date, datetime
 from typing import Iterable, NamedTuple, Optional, Tuple
 
 
@@ -46,6 +46,8 @@ class Amendement:
     observations: Optional[str] = None
     reponse: Optional[str] = None
 
+    bookmarked_at: Optional[datetime] = None
+
     @property
     def num_disp(self) -> str:
         text = str(self.num)
@@ -57,6 +59,10 @@ class Amendement:
             text += " "
             text += self._RECTIF_TO_SUFFIX[self.rectif]
         return text
+
+    @property
+    def num_str(self) -> str:
+        return str(self.num)
 
     @property
     def subdiv_disp(self) -> str:
--- a/repondeur/zam_repondeur/fetch/models.py
+++ b/repondeur/zam_repondeur/fetch/models.py
@@ -134,8 +134,14 @@ class Amendement:
         }
 
     @property
-    def url_key(self) -> str:
+    def lecture_url_key(self) -> str:
         return f"{self.chambre}.{self.session}.{self.num_texte}.{self.organe}"
+
+    @property
+    def article_url_key(self) -> str:
+        return (
+            f"{self.subdiv_type}.{self.subdiv_num}.{self.subdiv_mult}.{self.subdiv_pos}"
+        )
 
 
 class SubDiv(NamedTuple):
--- a/repondeur/zam_repondeur/fetch/models.py
+++ b/repondeur/zam_repondeur/fetch/models.py
@@ -133,6 +133,10 @@ class Amendement:
             if getattr(self, field_name) != getattr(other, field_name)
         }
 
+    @property
+    def url_key(self) -> str:
+        return f"{self.chambre}.{self.session}.{self.num_texte}.{self.organe}"
+
 
 class SubDiv(NamedTuple):
     type_: str
--- a/repondeur/zam_repondeur/resources.py
+++ b/repondeur/zam_repondeur/resources.py
@@ -1,4 +1,4 @@
-from typing import Iterator, List, Optional
+from typing import Iterator, List, Optional, cast
 
 from pyramid.request import Request
 
@@ -7,6 +7,10 @@ from zam_repondeur.models import (
     Lecture as LectureModel,
     DBSession,
 )
+
+
+class ResourceNotFound(Exception):
+    pass
 
 
 class Resource(dict):
@@ -21,7 +25,7 @@ class Resource(dict):
 
     @property
     def breadcrumbs_title(self) -> str:
-        raise NotImplementedError
+        raise NotImplementedError  # pragma: no cover
 
     breadcrumbs_link: bool = True
 
@@ -30,11 +34,15 @@ class Resource(dict):
         self.__parent__ = parent
 
     @property
+    def parent(self) -> Optional["Resource"]:
+        return self.__parent__
+
+    @property
     def parents(self) -> Iterator["Resource"]:
-        parent = self.__parent__
+        parent = self.parent
         while parent is not None:
             yield parent
-            parent = parent.__parent__
+            parent = parent.parent
 
     @property
     def ancestors(self) -> List["Resource"]:
@@ -90,7 +98,12 @@ class LectureResource(Resource):
         self.add_child(ArticleCollection(name="articles", parent=self))
 
     def model(self) -> LectureModel:
-        return LectureModel.get(self.chambre, self.session, self.num_texte, self.organe)
+        lecture = LectureModel.get(
+            self.chambre, self.session, self.num_texte, self.organe
+        )
+        if lecture is None:
+            raise ResourceNotFound(self)
+        return lecture
 
     @property
     def breadcrumbs_title(self) -> str:
@@ -104,6 +117,10 @@ class AmendementCollection(Resource):
     def __getitem__(self, key: str) -> Resource:
         return AmendementResource(name=key, parent=self)
 
+    @property
+    def parent(self) -> LectureResource:
+        return cast(LectureResource, self.__parent__)
+
 
 class AmendementResource(Resource):
     def __init__(self, name: str, parent: Resource) -> None:
@@ -111,11 +128,15 @@ class AmendementResource(Resource):
         self.num = int(name)
 
     @property
-    def lecture_resource(self) -> LectureResource:
-        return self.__parent__.__parent__  # type: ignore
+    def parent(self) -> AmendementCollection:
+        return cast(AmendementCollection, self.__parent__)
 
-    def model(self) -> Optional[AmendementModel]:
-        return (  # type: ignore
+    @property
+    def lecture_resource(self) -> LectureResource:
+        return self.parent.parent
+
+    def model(self) -> AmendementModel:
+        amendement: Optional[AmendementModel] = (
             DBSession.query(AmendementModel)
             .filter_by(
                 chambre=self.lecture_resource.chambre,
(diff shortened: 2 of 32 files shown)


repondeur/zam_repondeur/fetch/models.py
```diff
@@ -129,6 +129,16 @@ def changes(self, other: "Amendement", ignored_fields: Iterable[str] = ()) -> di
             if getattr(self, field_name) != getattr(other, field_name)
         }
 
+    @property
+    def lecture_url_key(self) -> str:
+        return f"{self.chambre}.{self.session}.{self.num_texte}.{self.organe}"
```
davidbgk: I like that new syntax 👍

repondeur/zam_repondeur/resources.py
```diff
@@ -0,0 +1,179 @@
...
+    Lecture as LectureModel,
+    DBSession,
+)
+
+
+class Resource(dict):
+    """Location-aware resource"""
+
+    __name__: Optional[str] = None
+    __parent__: Optional["Resource"] = None
```
davidbgk: Why `__parent__` over `parent`? We use it in templates so it doesn't look private nor magic to me.

repondeur/zam_repondeur/resources.py
```diff
@@ -0,0 +1,179 @@
...
+
+    @property
+    def breadcrumbs_title(self) -> str:
+        raise NotImplementedError
+
+    breadcrumbs_link: bool = True
+
+    def __init__(self, name: str, parent: "Resource") -> None:
```
davidbgk: I wonder if a `dataclass` would work here.

repondeur/zam_repondeur/resources.py
```diff
@@ -0,0 +1,179 @@
...
+    ) -> None:
+        super().__init__(name=name, parent=parent)
+        self.chambre = chambre
+        self.session = session
+        self.num_texte = num_texte
+        self.organe = organe
+        self.add_child(AmendementCollection(name="amendements", parent=self))
+        self.add_child(ArticleCollection(name="articles", parent=self))
+
+    def model(self) -> LectureModel:
```
davidbgk: Not fond of `models/model`, proposals: `list/item`? `collection/resource`? (redundant with the name but given the context that might more explicit). Why not properties?

repondeur/zam_repondeur/resources.py
```diff
@@ -0,0 +1,179 @@
...
+                organe=self.lecture_resource.organe,
+                num=self.num,
+            )
+            .first()
+        )
+
+    @property
+    def breadcrumbs_title(self) -> str:
+        amendement = self.model()
+        assert amendement is not None  # typing hint for mypy
```
davidbgk: Don't we want to raise if that's the case?

repondeur/zam_repondeur/resources.py
ronnix: That's the protocol used by Pyramid: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/resources.html#location-aware-resources

I think they used reserved attribute names so that it will not clash with existing attributes you might have on your resources.

repondeur/zam_repondeur/views/lectures.py
```diff
@@ -135,19 +130,17 @@ def post(self) -> Response:
...
+        self.context = context
         self.request = request
-        self.lecture = LectureModel.get(
-            chambre=request.matchdict["chambre"],
-            session=request.matchdict["session"],
-            num_texte=int(request.matchdict["num_texte"]),
-            organe=request.matchdict["organe"],
-        )
+        lecture_resource = cast(Optional[LectureResource], context.__parent__)
+        assert lecture_resource is not None
```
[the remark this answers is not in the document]
ronnix: That assert was just to appease mypy, as it cannot ever be None here, but I'll see if I can improve that.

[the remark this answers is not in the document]
ronnix: Looks like we can just do `lecture_resource  = cast(LectureResource, context.__parent__)` and remove the `assert`. 👍

repondeur/zam_repondeur/views/lectures.py
```diff
@@ -135,19 +130,17 @@ def post(self) -> Response:
...
+    def __init__(self, context: AmendementCollection, request: Request) -> None:
+        self.context = context
         self.request = request
-        self.lecture = LectureModel.get(
-            chambre=request.matchdict["chambre"],
-            session=request.matchdict["session"],
-            num_texte=int(request.matchdict["num_texte"]),
-            organe=request.matchdict["organe"],
-        )
+        lecture_resource = cast(Optional[LectureResource], context.__parent__)
```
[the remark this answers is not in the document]
ronnix: The `cast()` is just for the static type analysis, it does not change anything at runtime.

It tells mypy about the type of an expression. A typical use would be when mypy can infer a type  that is too general, and you know the function will return a more specific subtype.

If you write `a = cast(A, expr)`, you don't say anything about the `a` variable's type, so mypy will infer it based on the type of the assigned expression, which is `A` because of the cast.

If you write `a: A = expr`, it tells mypy about the type of the variable `a`. It will then infer the type of the expression and check that it's compatible with the type of the variable. If the infered type of `expr` is not `A` or a subtype of `A`, you'll get a type error.

I'll add a comment to the code to make it clear what's happening here.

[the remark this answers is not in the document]
ronnix: (At least that's my understanding, but I'm still new to this so I may be mistaken about how it works exactly...)

repondeur/zam_repondeur/views/reponse.py
```diff
@@ -1,24 +1,25 @@
...
 from pyramid.request import Request
 from pyramid.response import Response
 from pyramid.view import view_config, view_defaults
 from sqlalchemy.sql.expression import case
 
 from zam_repondeur.clean import clean_html
-from zam_repondeur.models import DBSession, Amendement as AmendementModel, AVIS, Lecture
+from zam_repondeur.models import (
+    DBSession,
+    Amendement as AmendementModel,
```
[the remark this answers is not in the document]
ronnix: You're right, I'm removing the aliases.

repondeur/zam_repondeur/views/reponse.py
```diff
@@ -53,50 +48,28 @@ def list_reponses(request: Request) -> Response:
...
+        if self.lecture is None:
+            raise HTTPNotFound
 
     @view_config(request_method="GET")
     def get(self) -> dict:
         return {"lecture": self.lecture, "amendement": self.amendement, "avis": AVIS}
 
     @view_config(request_method="POST")
     def post(self) -> Response:
+        assert self.amendement is not None  # typing hint for mypy
```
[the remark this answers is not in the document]
ronnix: Not enough for mypy.

I see an opportunity to make code simpler here. I think I'll do a refactoring to get rid of the `Optional`s and make sure the `model()` method never returns `None`, and raises something like a `ResourceNotFound` exception instead, for which we can add a generic 404 error view.

repondeur/zam_repondeur/resources.py
ronnix: Why not. Any benefits you think we could get by using a dataclass here?

repondeur/zam_repondeur/resources.py
ronnix: Sorry, I don't get it. Can you explain a bit more what you mean here?

repondeur/zam_repondeur/resources.py
ronnix: Yes, I'll make it raise when the model is not found.

repondeur/zam_repondeur/resources.py
ronnix: Oh, got it, nevermind!

repondeur/zam_repondeur/resources.py
ronnix: Anyway, I'll add a `parent` property so that we don't have to access `__parent__` directly, and to make overriding the type in subclasses easier.

repondeur/zam_repondeur/views/reponse.py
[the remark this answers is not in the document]
ronnix: Done in 7a06415

repondeur/zam_repondeur/templates/article_edit.html
```diff
@@ -44,7 +36,7 @@ <h2>{{ amendement.subdiv_disp }}</h2>
         </div>
         <div class="form-group">
             <input type="submit" name="submit" value="Enregistrer" class="btn btn-primary">
-            <a href="{{ request.route_url('list_amendements', chambre=amendement.chambre, session=amendement.session, num_texte=amendement.num_texte, organe=amendement.organe) }}" class="btn">Annuler</a>
+            <a href="{{ request.resource_url(context.__parent__.__parent__['amendements']) }}" class="btn">Annuler</a>
```
davidbgk: Could be `context.parent.parent['amendements']` if I understand the latest changes correctly?

repondeur/zam_repondeur/templates/reponse_edit.html
```diff
@@ -116,7 +108,7 @@ <h4>Réponse</h4>
 
         <div class="form-group">
             <input type="submit" name="submit" value="Enregistrer" class="btn btn-primary">
-            <a href="{{ request.route_url('list_amendements', chambre=amendement.chambre, session=amendement.session, num_texte=amendement.num_texte, organe=amendement.organe) }}" class="btn">Annuler</a>
+            <a href="{{ request.resource_url(context.__parent__) }}" class="btn">Annuler</a>
```
davidbgk: Same here?

repondeur/zam_repondeur/writer.py
```diff
@@ -36,7 +36,7 @@ def rename_field(field_name: str) -> str:
     return FIELDS_NAMES.get(field_name, field_name.capitalize())
 
 
-EXCLUDED_FIELDS = ["subdiv_contenu"]
+EXCLUDED_FIELDS = ["subdiv_contenu", "bookmarked_at"]
```
davidbgk: 👍

repondeur/zam_repondeur/templates/article_edit.html
ronnix: Thought I converted them all but looks like some have slipped through!

repondeur/zam_repondeur/resources.py
```diff
@@ -0,0 +1,215 @@
...
+        lecture = LectureModel.get(
+            self.chambre, self.session, self.num_texte, self.organe
+        )
+        if lecture is None:
+            raise ResourceNotFound(self)
+        return lecture
+
+    @property
+    def breadcrumbs_title(self) -> str:
+        return str(self.model())
```
davidbgk: Proposal: `return f"{self.model().dossier_legislatif} → {str(self.model())}"` otherwise we are loosing the information related to the text.

ronnix: Thanks, I missed that when rebasing!

PR #90, opened 2018-07-24
Dealing with sous-amendements

davidbgk:


repondeur/zam_repondeur/models/amendement.py
```diff
@@ -75,6 +75,7 @@
     Column("position", Integer, nullable=True),
     Column("discussion_commune", Integer, nullable=True),
     Column("identique", Boolean, nullable=True),
+    Column("parent", Text, nullable=True),
```
ronnix: `num` est un `Integer` donc je me dis que ça aurait du sens d'utiliser ce même type pour `parent` ?

repondeur/zam_repondeur/writer.py
```diff
@@ -36,7 +36,7 @@ def rename_field(field_name: str) -> str:
     return FIELDS_NAMES.get(field_name, field_name.capitalize())
 
 
-EXCLUDED_FIELDS = ["subdiv_contenu", "bookmarked_at"]
+EXCLUDED_FIELDS = ["subdiv_contenu", "bookmarked_at", "parent"]
```
ronnix: Peut-être que c'est une info valable à exporter ?

repondeur/zam_repondeur/models/amendement.py
davidbgk: C'est traitre car ça peut être `128 (Rect)` par exemple dans ce champ, le top ça serait de pouvoir faire un lien vers l'amendement à terme.

ronnix:
Also, maybe it's time to add database migrations, so we don't have to wipe the DB every time we change the schema...

repondeur/zam_repondeur/models/amendement.py
ronnix: Dans ce cas là, on voudrait juste garder `128`. Peut-être qu'on peut utiliser `Amendement.parse_num()` ?

ronnix:
Added migrations in #92 so when it's merged I think we can add a migration on this PR to add the extra column. (Sqlite has some limitations with schema changes, like renaming a column is unnecessarily complicated, but hopefully we won't run into that here...)

repondeur/zam_repondeur/fetch/models.py
```diff
@@ -48,6 +50,20 @@ class Amendement:
 
     bookmarked_at: Optional[datetime] = None
 
+    @property
+    def parent(self) -> Optional["Amendement"]:
```
davidbgk: @ronnix I'm pretty sure SQLAlchemy should do that for us, thoughts?

davidbgk:
@ronnix r?

ronnix:
Why do we need to store the `rectif` field ?

davidbgk:
> Why do we need to store the `rectif` field ?

Mostly to display/query it correctly.

ronnix:
Display I can understand, and maybe it’s a good reason to keep it.

I’m not sure about the query part though. See my question on Slack. If a parent amendement gets modified (its rectif increases), I think we should still consider the child amendement to be related to it. What do you think?

davidbgk:
It's hard to choose without a real case and we'll adapt given available data, about the relation with self-reference, do you think we should keep it as is or use some kind of `relation`?

PR #98, opened 2018-07-30
Add the comments field for amendement reponse

davidbgk:


ronnix:
Maybe update the `test_post_form` test to make sure the field is saved?

PR #106, opened 2018-08-20
Introduce asynchronous tasks through Huey

davidbgk:


repondeur/zam_repondeur/huey_launcher.py
```diff
@@ -0,0 +1,24 @@
...
+    "zam.legislature": "15",
+    "zam.secret": "dummy",
+    "zam.an_groups_folder": HERE.parent / "data" / "an" / "groups" / "organe",
+}
+
+with Configurator(settings=settings, root_factory=Root) as config:
+    config.include("pyramid_tm")
+    engine = engine_from_config(settings, "sqlalchemy.")
+    DBSession.configure(bind=engine)
+    Base.metadata.bind = engine
```
ronnix: This one feels a bit weird. We use import-time side effects to initialize our stuff when the huey_consumer script imports the huey instance. Maybe we can reverse it, and use our own script as the entry point, that would set up the pyramid context (
https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/commandline.html#writing-a-scrip)t, then do the 2 important lines from 
https://github.com/coleifer/huey/blob/master/huey/bin/huey_consumer.py#L46-L47 ?

PR #109, opened 2018-09-10
Fix parsing of Assemblée nationale amendement numbers for commissions

ronnix:
We only handled the "COM-" prefix, but it looks like others can be used, so we'll just ignore any leading uppercase letters, "-" and "|".

Fixes: https://rollbar.com/zam/zam/items/25/

### DIFF
--- a/repondeur/tests/models/test_amendement.py
+++ b/repondeur/tests/models/test_amendement.py
@@ -4,36 +4,27 @@ import pytest
 from sqlalchemy.exc import IntegrityError
 
 EXAMPLES = [
-    ("42", 42, 0),
-    ("42 rect.", 42, 1),
-    ("42 rect. bis", 42, 2),
-    ("42 rect. ter", 42, 3),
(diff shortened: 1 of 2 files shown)


boogheta:
Hi!
If I may, this seems like a dangerous choice IMHO: it happens quite often multiple commissions are studying a same texte, so you will end up with multiple amendments with the same number on a same text.
In NosDéputés, we always keep the text identifying the commitee as a prefix of the number.

ronnix:
Hi @boogheta,

Thanks for your comment!

We're aware of this, and in addition to the amendement number we also store the related "lecture" (séance publique, commission...) for this text, and we have a unicity constraint on the `lecture` and `num` pair (see [here](https://github.com/betagouv/zam/blob/master/repondeur/zam_repondeur/models/amendement.py#L33)).

Does this alleviate your worries?

davidbgk:
👍

PR #110, opened 2018-09-10
Fix updating amendements from Assemblée nationale if article changes

ronnix:
When the article has changed since we first created the amendement, we would fail to find it again, then try to insert it again, causing an `IntegrityError` as the unicity constraint in the database fails.

Fixes: https://rollbar.com/zam/zam/items/26/


davidbgk:
👍

PR #111, opened 2018-09-10
Fix incomplete fetching of sous-amendements

ronnix:
When fetching amendements from Assemblée nationale, only the last sous-amendement would have its parent set correctly.

The issue seems to be due to how the SQLAlchemy relationship was configured + not doing a flush after setting the parent amendement, which would cause the changed not to be fully registered (not sure exactly why).


repondeur/zam_repondeur/models/amendement.py
```diff
@@ -57,7 +57,9 @@ class Amendement(Base):  # type: ignore
     # Relations.
     parent_pk = Column(Integer, ForeignKey("amendements.pk"), nullable=True)
     parent_rectif = Column(Integer, nullable=True)
-    parent = relationship("Amendement", uselist=False, primaryjoin=(parent_pk == pk))
+    parent = relationship(
+        "Amendement", uselist=False, remote_side=[pk], backref=backref("children")
```
davidbgk: TODO(self): check if that change doesn't increase the number of queries to display the list of amendements and the visionneuse, even better: set a test for that 😇

PR #112, opened 2018-09-11
Add a confirmation message after user updates a response

ronnix:


davidbgk:
I don't really get the reason of the named tuple but why not.

👍

PR #117, opened 2018-09-11
Fix circular refs when deleting lecture with sous-amendements

ronnix:
For some mysterious reason, some lectures cannot be deleted because of circular references between amendements.

This should not happen, and I have not been able to reproduce it yet, but this fix should prevent the error from happening again.

See: http://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.post_update


davidbgk:
😨 👍

PR #123, opened 2018-09-18
Add hash-based query strings to static assets for cache busting

ronnix:


repondeur/setup.cfg
```diff
@@ -17,12 +17,12 @@ show_missing = True
 [mypy]
 ignore_missing_imports = True
 
-; strict mode
+; strict mode (almost)
```
davidbgk: 😅

PR #126, opened 2018-09-18
Implement new design

ronnix:
WIP : we should refactor and remove bootstrap


repondeur/zam_repondeur/templates/amendements.html
```diff
@@ -1,152 +1,250 @@
...
+
+    /* Sorting */
+    .headers th {
+        cursor: pointer;
+    }
+    .headers th.nosort {
+        cursor: default;
+    }
+    .headers th::before {
+        font-family: Font Awesome\ 5 Free;
```
davidbgk: I think that better than escaping the space before the 5 is to double quote the font name?

repondeur/zam_repondeur/templates/amendements.html
```diff
@@ -1,152 +1,250 @@
...
+    <table class="table" style="width:100%">
+        <thead class="box">
+            <tr class="headers">
+                <th class="nosort"></th>
+                <th width="15%">Article</th>
+                <th width="15%">Nº</th>
                 <th>Auteur</th>
                 <th>Groupe</th>
-                <th>Sort</th>
                 <th>
```
davidbgk: Maybe we should force the `width` here too to avoid the moving parts when we open the advanced filters?

repondeur/zam_repondeur/templates/amendements.html
ronnix: Yes. I think I copied it from a minified version (1 char less!!).

repondeur/zam_repondeur/templates/amendements.html
ronnix: Sounds good.

PR #130, opened 2018-09-19
Retrieve annexes content

davidbgk:
Requires https://github.com/regardscitoyens/the-law-factory-parser/issues/103 to be solved though.

### DIFF
--- a/repondeur/zam_repondeur/fetch/__init__.py
+++ b/repondeur/zam_repondeur/fetch/__init__.py
@@ -120,7 +120,7 @@ def get_possible_texte_urls(lecture: Lecture) -> List[str]:
 
 def parse_first_working_url(urls: List[str]) -> List[dict]:
     for url in urls:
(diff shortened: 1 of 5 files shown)


repondeur/zam_repondeur/fetch/__init__.py
```diff
@@ -71,6 +71,10 @@ def get_article_num_mult(title: str) -> Tuple[str, str]:
     title = title.replace("1er", "1").replace("liminaire", "0")
     if " " in title:
         num, mult = title.split(" ", 1)
+        try:
+            int(num)
+        except ValueError:
+            return "", ""
```
davidbgk: This is a problem in case there are many annexes given the following `get_one_or_create` will be unable to make the distinction (same empty num/mult), any idea?

ronnix:
@davidbgk J'ai rebasé et squashé, résolu les conflits, tout ça, dis-moi si ça te semble correct !

PR #131, opened 2018-09-19
Scroll to the updated amendement once saved

davidbgk:


repondeur/zam_repondeur/models/amendement.py
```diff
@@ -140,8 +144,8 @@ def num_disp(self) -> str:
         return text
 
     @property
-    def num_str(self) -> str:
-        return str(self.num)
+    def slug(self):
```
ronnix: Missing type annotation?

repondeur/zam_repondeur/views/reponse.py
```diff
@@ -56,4 +56,5 @@ def post(self) -> Response:
         )
 
         collection: AmendementCollection = self.context.parent
-        return HTTPFound(location=self.request.resource_url(collection))
+        url = self.request.resource_url(collection)
```
ronnix: You should be able to use resource_url(collection, anchor=self.amendement.slug)

PR #145, opened 2018-09-25
Handle not found resources as 404

davidbgk:


ronnix:
Mmm... On a ça qui aurait dû faire le taf, du moins c’était l’intention : 
https://github.com/betagouv/zam/blob/99e5e03973aef036a5d772af14af3a0b2669c2e5/repondeur/zam_repondeur/views/errors.py

davidbgk:
Ah, pas vu, peut-être pas importé ?

ronnix:
Normalement le config.scan() explore tout le projet, sans que tu aies besoin d’importer toutes tes vues explicitement. C’est bizarre.

ronnix:
https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/views.html#custom-exception-views

ronnix:
I think I understand what's happening.

The `ResourceNotFound` exception is actually caught by the exception view tween, and we do return a 404. But the exception also goes through the rollbar tween before being handled, so we get a report. :-(

Subclassing from `HTTPNotFound` is thus a good idea. It will silence rollbar, as it ignores all instances of `WSGIHTTPException`.

PR #154, opened 2018-10-01
Fix fetching list of only 1 amendement from AN

ronnix:


repondeur/zam_repondeur/fetch/an/amendements.py
```diff
@@ -87,6 +87,9 @@ def fetch_amendements(lecture: Lecture) -> List[OrderedDict]:
     amendements_raw: List[OrderedDict] = (
         content["amdtsParOrdreDeDiscussion"].get("amendements", [])["amendement"]
     )
+    # If there is only 1 amendement, xmltodict does not return a list :(
```
davidbgk: Isn't the above line wrong in terms of annotations in this case?

ronnix: Good catch!

PR #155, opened 2018-10-02
Preserve case of article numbers (e.g. "Chapitre III")

ronnix:


ronnix:
Extracted it from #150

PR #159, opened 2018-10-02
Make sure different update scenarii are covered

ronnix:
This includes tests and fixes for fetching amendements from AN and Sénat.

TODO: do the same for imports, especially liasse XML.


ronnix:
Cool, je vais déjà merger ça !

PR #161, opened 2018-10-03
Fix ordering of articles with interstitial markers ("AAAAA")

ronnix:
https://fr.wikipedia.org/wiki/Association_amicale_des_amateurs_d%27andouillette_authentique

### DIFF
--- a/repondeur/tests/models/test_article.py
+++ b/repondeur/tests/models/test_article.py
@@ -1,6 +1,17 @@
+from itertools import tee
+
 import pytest
 
 from zam_repondeur.models import Article
+
+
+def pairwise(iterable):
+    """
+    From https://docs.python.org/3/library/itertools.html#itertools-recipes
+    """
+    a, b = tee(iterable)
+    next(b, None)
+    return zip(a, b)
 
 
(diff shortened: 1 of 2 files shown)


ronnix:
Should fix https://rollbar.com/zam/zam/items/51/ and https://rollbar.com/zam/zam/items/52/

repondeur/tests/models/test_article.py
```diff
@@ -77,6 +88,72 @@ def test_types(self, lecture_an):
...
+                    "17 bis BB",
+                    "17 bis B",
+                    "17 bis C",
+                    "17 bis",
+                    "17 ter",
+                ]
+            )
+        ),
+    )
+    def test_andouillette(self, lecture_an, before, after):
```
davidbgk: ❤️ can't wait for that one to fail, it deserves a dedicated pytest plugin!

ronnix: 😋

boogheta:
You should really explore first the pieces of code of tlfp ;)
https://github.com/regardscitoyens/the-law-factory-parser/blob/master/tlfp/tools/sort_articles.py

PR #162, opened 2018-10-03
Add unique constraint on articles

ronnix:


repondeur/zam_repondeur/models/article.py
```diff
@@ -53,6 +54,8 @@ class Article(Base):
     contenu = Column(PickleType, nullable=True)
     jaune = Column(Text, nullable=True)  # Présentation de l’article.
 
+    __table_args__ = (UniqueConstraint("lecture_pk", "type", "num", "mult", "pos"),)
```
davidbgk: We put that line under `__tablename__`  for `Amendement`, not sure if it's relevant but more consistent :)

ronnix: Right

PR #165, opened 2018-10-03
Worker improvements

ronnix:
Multiple changes conspiring together to avoid the situation where an initial fetch is blocked behind a queue of periodic updates...

### DIFF
--- a/repondeur/zam_repondeur/tasks/fetch.py
+++ b/repondeur/zam_repondeur/tasks/fetch.py
@@ -14,7 +14,10 @@ from zam_repondeur.models import DBSession, Journal, Lecture
 logger = logging.getLogger(__name__)
 
 
-@huey.task(retries=3, retry_delay=60 * 5)  # Minutes.
+RETRY_DELAY = 5 * 60  # 5 minutes
+
+
+@huey.task(retries=3, retry_delay=RETRY_DELAY)
 @huey.lock_task("fetch-articles-lock")
 def fetch_articles(lecture_pk: int) -> None:
     with transaction.manager:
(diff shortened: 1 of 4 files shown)


repondeur/zam_repondeur/tasks/fetch.py
```diff
@@ -14,10 +14,12 @@
...
 
-@huey.task(retries=3, retry_delay=60 * 5)  # Minutes.
-@huey.lock_task("fetch-articles-lock")
+RETRY_DELAY = 5 * 60  # 5 minutes
+
+
+@huey.task(retries=3, retry_delay=RETRY_DELAY)
 def fetch_articles(lecture_pk: int) -> None:
-    with transaction.manager:
+    with transaction.manager, huey.lock_task(f"fetch-articles-{lecture_pk}"):
```
davidbgk: 👌

repondeur/zam_repondeur/tasks/periodic.py
```diff
@@ -23,4 +23,5 @@ def update_data() -> None:
 def fetch_all_amendements() -> None:
     with transaction.manager:
         for lecture in DBSession.query(Lecture):
-            fetch_amendements(lecture.pk)
+            delay = lecture.pk % (15 * 60)  # spread out updates over 15 minutes
```
davidbgk: I'm not sure we should rely on the lecture's pk, maybe safer to enumerate on lectures and compute the delay based on an truly incremental index?

ronnix: The idea was that using the `lecture_pk` it allows each lecture to be updated every hour, with a stable offset. What do you think ?

ronnix:
It's more work to make the periodic task configurable, because it's defined at import time in the decorator... ☹️

davidbgk:
[Indeed](https://huey.readthedocs.io/en/latest/api.html#Huey.periodic_task), maybe coding our custom [crontab](https://huey.readthedocs.io/en/latest/api.html#crontab) method which relies on settings but I'm not even sure we can delay that 🤔

repondeur/zam_repondeur/tasks/periodic.py
ronnix: I guess I'm tired 🙂

PR #166, opened 2018-10-03
Ability to export a PDF with a unique amendement

davidbgk:


admin/nginx-https.conf.template
```diff
@@ -45,13 +45,13 @@ server {
         proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header        X-Forwarded-Proto $scheme;
 
-        client_max_body_size    10m;
+        client_max_body_size    20m;
```
ronnix: On veut accepter des gros uploads ?

davidbgk: *client* tabarouette /selfslap

PR #175, opened 2018-10-08
Redirect to répondeur if no changelog has been deployed

ronnix:


admin/nginx-https.conf.template
```diff
@@ -26,16 +26,20 @@ server {
...
+    auth_basic "Restricted";
+    auth_basic_user_file /etc/nginx/.htpasswd;
+
+    # Redirect to repondeur if there is no changelog
+    location = / {
+        alias /srv/zam/;
+        try_files index.html /repondeur/;
+    }
+
     location / {
```
davidbgk: Is that entry still relevant?

ronnix: There are some let’s encrypt related files there, I supposed we were supposed to serve them. What do you think?

davidbgk: Oh, I missed the equal sign above, my bad.

ronnix: No, nginx config format is a bit esoteric...

PR #189, opened 2018-10-11
Explicit save and back actions on reponse edit + go to next article on jaune edition

davidbgk:
The back button no longer needs the anchor/hash given that we don't want to highlight the line when we are coming back BUT keep the highlight when we save the response.

### DIFF
--- a/repondeur/zam_repondeur/models/article.py
+++ b/repondeur/zam_repondeur/models/article.py
@@ -200,19 +200,19 @@ class Article(Base):
         return article
 
     @property
-    def previous_article(self) -> Union["Article", bool]:
+    def previous_article(self) -> Optional["Article"]:
         sorted_articles: List[Article] = sorted(self.lecture.articles)
         previous_index = sorted_articles.index(self) - 1
         if previous_index < 0:
-            return False
+            return None
         return sorted_articles[previous_index]
 
     @property
-    def next_article(self) -> Union["Article", bool]:
+    def next_article(self) -> Optional["Article"]:
         sorted_articles: List[Article] = sorted(self.lecture.articles)
         next_index = sorted_articles.index(self) + 1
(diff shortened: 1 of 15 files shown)


ronnix:
We dont want the highlight in that case, but it would be nice to go back to our position in the list 🤔

repondeur/zam_repondeur/templates/amendements.html
```diff
@@ -763,6 +770,7 @@ <h3>Supprimer</h3>
         editLink.addEventListener('click', e => {
             const thisURL = new URL(window.location.href)
             const linkURL = new URL(e.target.href)
+            thisURL.hash = ''
```
ronnix: The hash is not included in the value of `back` below, so I don't think this is needed

repondeur/zam_repondeur/models/article.py
```diff
@@ -199,6 +199,22 @@ def create(
...
+        return sorted_articles[previous_index]
+
+    @property
+    def next_article(self) -> Optional["Article"]:
+        sorted_articles: List[Article] = sorted(self.lecture.articles)
+        next_index = sorted_articles.index(self) + 1
+        if next_index >= len(sorted_articles):
+            return None
+        return sorted_articles[next_index]
+
```
ronnix: Maybe we should skip / filter out intersticial articles. We don't show them in the list as we don't need to edit info about them, so it's surprising to land on one when navigating to the next article.

davidbgk: It is done if you want to take a look

repondeur/zam_repondeur/templates/amendements.html
davidbgk: I think there is a tricky case when there is already a remaining hash coming from a previous back/save action which keep it. I'll double check

PR #191, opened 2018-10-12
Use an official wkhtmltopdf build

ronnix:
The build in the distro is not the latest, but mostly it does not include some Qt patches, causing some issues:
- slow PDF generation
- excessively large file size
- missing footers and table of contents
- impossibility to select text


ronnix:
I don't see an easy way to do that. But may I suggest the opposite: using Docker in production? 😛

PR #201, opened 2018-10-18
Ignore unknown groupeTribunId when fetching AN amendements

ronnix:


repondeur/zam_repondeur/fetch/an/amendements.py
```diff
@@ -219,9 +219,17 @@ def get_groupe(amendement: OrderedDict) -> str:
...
+    try:
+        groupe_tribun_id = f"PO{auteur['groupeTribunId']}"
+    except KeyError:
+        logger.error(
+            "Unknown groupe %r for amendement %s",
+            groupe_tribun_id,
+            amendement["numero"],
+        )
+        return ""
+    groupe: Dict[str, str] = groupes[groupe_tribun_id]
```
davidbgk: Are we sure that line and the following one will not raise a `KeyError` at some point too?

ronnix: We've never met this case... yet. 😉

boogheta:
FYI, what often happens is that the TribunId is in fact the ID of a commitee, for instance when a "rapporteur pour avis" proposes in hemicycle amendments adopted by the commitee

PR #212, opened 2018-10-23
Improve PDFs exports

davidbgk:
1. add tests
2. restore sous-amendements after parent ones
3. remove (empty) additional article pages


ronnix:
👍

PR #214, opened 2018-10-24
Allow child amendements with missing (but known) parents in a liasse XML

ronnix:


repondeur/zam_repondeur/views/liasse.py
```diff
@@ -62,6 +62,16 @@ def _do_upload_liasse_xml(context: LectureResource, request: Request) -> Respons
...
         return
 
+    if errors:
+        if len(errors) == 1:
+            what = "l'amendement"
+        else:
+            what = "les amendements"
+        details = ", ".join(f"{uid} ({cause})" for uid, cause in errors)
+        request.session.flash(
+            Message(cls="warning", text=f"Impossible d'importer {what} {details}.")
```
davidbgk: This message will mix French and English, probably not in a really meaningful way for the user. Maybe display the uids only to help resolution and post the cause to rollbar for us?

repondeur/zam_repondeur/fetch/an/liasse_xml.py
```diff
@@ -47,18 +49,22 @@ def import_liasse_xml(xml_file: IO[bytes]) -> List[Amendement]:
...
 
         uid = child.find(f"./{NS}uid").text
-        amendement = _make_amendement(child, uid_map)
-        uid_map[uid] = amendement
+        try:
+            amendement = _make_amendement(child, uid_map)
+            uid_map[uid] = amendement
+        except Exception as exc:
+            logger.exception(f"Failed to import amendement {uid} from liasse")
+            errors.append((uid, str(exc)))
```
davidbgk: Do we still need the exception trace in errors?

ronnix: I think it can help us understand what went wrong exactly.

What do you think?

davidbgk: Alright, I don't get where it is actually sent from here?

ronnix: See b7f3107

PR #221, opened 2018-10-31
Improvements on amendements list

davidbgk:


davidbgk:
I hesitated to remove it :)

My main point is to keep the data for a while if there is a rollback at some point and clean up later. Thoughts?

ronnix:
I think that’s fine. Let’s remove the feature and keep the data while we get feedback.

PR #242, opened 2018-11-08
Include « Objet de l'amendement » field for governement amendements too

ronnix:


ronnix:
Users don't know what they want and that's okay. Let's test different things together until we're all happy with what we have! 😃

PR #249, opened 2018-11-19
Show groups of identical amendements in répondeur

davidbgk:


repondeur/zam_repondeur/writer.py
```diff
@@ -26,8 +26,8 @@
     "num": "Num amdt",
     "auteur": "Auteur(s)",
     "date_depot": "Date de dépôt",
-    "discussion_commune": "Discussion commune ?",
-    "identique": "Identique ?",
+    "id_discussion_commune": "Discussion commune ?",
```
davidbgk: I think we should change these two labels.

ronnix: Ideas?

davidbgk: Identifiant discussion commune / Identifiant identique ?

PR #250, opened 2018-11-21
First iteration on adding tests for JS frontend

davidbgk:
Still requires documentation, to test you must have the geckodriver in your bin folder https://github.com/mozilla/geckodriver/releases

### DIFF
--- a/repondeur/tests/integration/conftest.py
+++ b/repondeur/tests/integration/conftest.py
@@ -6,11 +6,8 @@ from webtest.http import StopableWSGIServer
 
 @pytest.fixture(scope="session")
 def driver():
-    options = webdriver.firefox.options.Options()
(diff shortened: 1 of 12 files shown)


repondeur/tests/integration/conftest.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from selenium import webdriver
+from selenium.common.exceptions import WebDriverException
+from webtest.http import StopableWSGIServer
+
+
+@pytest.fixture#(scope="session")
```
davidbgk: There is definitely something wrong here for now, either because the browser is caching some pages (fragments?) or because the `DBSession` is not flushed prior the browser request is issued. Especially visible with `test_visionneuse_detail` serie. To be continued.

PR #263, opened 2018-12-10
Split Article and Amendement models

davidbgk:
First step before versioning.


repondeur/zam_repondeur/fetch/an/amendements.py
```diff
@@ -296,8 +294,8 @@ def _create_or_update_amendement(
         groupe = get_groupe(raw_auteur, amendement.num)
         auteur = get_auteur(raw_auteur)
 
-    dispositif = unjustify(get_str_or_none(amend, "dispositif") or "")
-    objet = unjustify(get_str_or_none(amend, "exposeSommaire") or "")
+    expose = unjustify(get_str_or_none(amend, "dispositif") or "")
+    corps = unjustify(get_str_or_none(amend, "exposeSommaire") or "")
```
davidbgk: I'm quite surprised that my chain of renaming leads to putting the `exposeSommaire` into the `corps` and not the `expose`.  There is another place where it was quite weird too, let me reference it too.

repondeur/zam_repondeur/writer.py
```diff
@@ -36,9 +36,9 @@
     "id_discussion_commune": "Identifiant discussion commune",
     "id_identique": "Identifiant identique",
     "parent": "Parent",
-    "dispositif": "Corps amdt",
-    "objet": "Exposé amdt",
-    "observations": "Objet amdt",
+    "corps": "Corps amdt",
+    "expose": "Exposé amdt",
```
davidbgk: Here is the second case where `objet` was previously labeled as `Exposé amdt`, we might have a switch on both side of the chain. It doesn't explain why it looks to be OK on the HTML side though 🤔

PR #264, opened 2018-12-12
POC of versionning using SQLAlchemy events

davidbgk:
DO NOT MERGE, let's discuss 🤗

Very basic but it gives a direction.


davidbgk:
See https://github.com/betagouv/zam/pull/289

PR #275, opened 2019-01-07
Install the Python "wheel" package to be able to build a wheel cache

ronnix:
This gets rid of the following errors when installing packages:

  error: invalid command 'bdist_wheel'


ronnix:
Closing in favour of #276

PR #276, opened 2019-01-07
Upgrade dependencies

ronnix:


boogheta:
Hey, it's mostly to ensure stability in the long term, we upgrade them manually from time to time, do you need some upgrades on our side?

davidbgk:
@boogheta the only error message I have so far is `lawfactory-utils 0.1.7 has requirement lxml<4.3,>=4.2, but you'll have lxml 4.3.0 which is incompatible.` and it doesn't look to be a blocker.

PR #278, opened 2019-01-08
Set basic users/teams models

davidbgk:


repondeur/zam_repondeur/models/users.py
```diff
@@ -0,0 +1,54 @@
...
+        DBSession.add(team)
+        return team
+
+
+class User(Base):
+    __tablename__ = "users"
+    __repr_keys__ = ("name", "email", "teams")
+
+    pk = Column(Integer, primary_key=True)
+    email = Column(String(254), nullable=False)
```
davidbgk: I wonder if something like https://sqlalchemy-utils.readthedocs.io/en/latest/data_types.html#module-sqlalchemy_utils.types.email would be pertinent 🤔

repondeur/zam_repondeur/models/users.py
```diff
@@ -0,0 +1,54 @@
...
+    name = Column(Text)
+    created_at = Column(
+        DateTime, nullable=False, default=datetime.utcnow, server_default=func.now()
```
davidbgk: Is https://github.com/betagouv/zam/commit/743af4925fa62237e1a10c3480c7deb20a072c46#r31721983 still relevant? 😇

repondeur/zam_repondeur/views/auth.py
```diff
@@ -21,7 +24,11 @@ def get(self) -> Any:
     @view_config(request_method="POST")
     def post(self) -> Any:
         email = self.request.params["email"].strip().lower()
-        headers = remember(self.request, email)
+        user, created = get_one_or_create(User, email=email)
+        user.last_login_at = datetime.utcnow()
+        if created:
```
davidbgk: I wonder if we should put that flush directly in `get_one_or_create`?

repondeur/zam_repondeur/__init__.py
```diff
@@ -116,3 +120,46 @@ def setup_auth(config: Configurator, settings: dict) -> None:
...
     config.add_route("logout", "/deconnexion")
+
+
+def get_user(request: Request) -> Optional[User]:
+    """
+    Find the user in the database based on the ID in the auth cookie
+    """
+    user_id = request.unauthenticated_userid
+    if user_id is not None:
+        user: Optional[User] = DBSession.query(User).get(user_id)
```
davidbgk: Don't we want to be a bit more defensive here? Like after a DB reset the `user_id` might not exist anymore?

repondeur/zam_repondeur/views/auth.py
ronnix: I don't think we should do that. (I'm not even sure I like doing the `add` in there...)

repondeur/zam_repondeur/models/users.py
ronnix: Yeah, we should probably do a clean up of our datetimes at some point.

repondeur/zam_repondeur/models/users.py
ronnix: That's a good idea.

repondeur/zam_repondeur/__init__.py
ronnix: In that case, `get` will return `None`, and the user will not be authenticated.

PR #285, opened 2019-01-16
Fix fetching of "rect." number from Assemblée nationale

ronnix:


repondeur/zam_repondeur/fetch/an/amendements.py
```diff
@@ -467,6 +468,27 @@ def get_parent_raw_num(amendement: OrderedDict) -> str:
...
+    (?P<rect>\s\((?:(?P<rect_mult>\d+)\w+\s)?Rect\))?
+    """,
+    re.VERBOSE,
+)
+
+
+def parse_numero_long_with_rect(text: str) -> int:
+    mo = RE_NUM_LONG.match(text)
+    if mo is not None and mo.group("rect"):
+        return int(mo.group("rect_mult") or 1)
```
davidbgk: I guess there will be a lot of weird cases (as always 🤷‍♂️ ), maybe a `elif "rect" in text: log-somewhere` would help to find these cases?

ronnix: I've looked at a lot of data and it looks pretty consistent.

PR #287, opened 2019-01-16
Synchronuous articles fetching on lecture creation

davidbgk:
Otherwise the async task is locked given that it shares the lock with async amendements fetching which leads to a 5 minutes delay between lecture creation and articles’ content retrieval.

The trade-off is to wait a bit on creation but articles retrieval is less time consuming (only one HTML page to scrape).


repondeur/zam_repondeur/views/lectures.py
```diff
@@ -81,14 +82,19 @@ def post(self) -> Response:
...
         # thus many objects within the database.
         transaction.commit()
+        get_articles(lecture_model)
+        Journal.create(
+            lecture=lecture_model,
+            kind="info",
+            message="Récupération des articles effectuée.",
+        )
         fetch_amendements(lecture_model.pk)
         fetch_articles(lecture_model.pk)
```
ronnix: Maybe we can remove this one, then?

repondeur/zam_repondeur/views/lectures.py
```diff
@@ -81,14 +82,19 @@ def post(self) -> Response:
         # for future access. Otherwise, it may create many instances and
         # thus many objects within the database.
         transaction.commit()
+        get_articles(lecture_model)
+        Journal.create(
+            lecture=lecture_model,
+            kind="info",
+            message="Récupération des articles effectuée.",
+        )
```
ronnix: We can probably move this block above the transaction.commit() and associated comment

PR #289, opened 2019-01-19
Events and journals

davidbgk:


davidbgk:
At some point we will want to remove 2 migrations but let's do that while pairing to update the revision number together :)

ronnix:
Trying to add the auto rebase label again (not sure why the bot removed it)

ronnix:
(My hope is that this way we don’t need to merge master into the branch, or rebase it manually)

davidbgk:
@ronnix there was a conflict so I think it's running fine :)

ronnix:
Ah so maybe it removes the label when some human intervention is needed? 🤔

davidbgk:
That would be (maybe too) clever!

ronnix:
Merci, petit robot !

PR #297, opened 2019-01-30
Optimize Sénat fetch

ronnix:
I looked into which tests were slow, and the worst offender was `tests/fetch/test_fetch_senat.py::test_aspire_senat_plf2019_1re_partie` with ~27s on my computer.

I then did some profiling to understand where time was being spent:
capture d ecran 2019-01-29 a 13 14 46

The biggest one is `get_one_or_create`, which does a `SELECT` query to the database each time, additionally causing an automatic session flush if we added some objects to the session.

I refactored the fetch logic to avoid doing queries and doing the article/amendement lookup Python-side by iterating on the relationship from the lecture object.

This brings a significant speed improvement:
capture d ecran 2019-01-30 a 16 55 52

Most of the time is then spent in the HTML cleaning.

I was able to reduce it a little by using a global `Cleaner` object to avoid paying the initialization cost on each call:
capture d ecran 2019-01-30 a 16 58 55

The remaining time is spent parsing the HTML with html5lib and serializing it again.

I tried using the `html-sanitizer` library instead of `bleach`, as it is based on `lxml`. The speed is better, but the behaviour is a bit different, so I preferred not to include that change in this PR.

### DIFF
--- a/repondeur/zam_repondeur/models/lecture.py
+++ b/repondeur/zam_repondeur/models/lecture.py
@@ -1,12 +1,16 @@
 from datetime import datetime
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 
 from sqlalchemy import Column, DateTime, Index, Integer, Text, desc
 from sqlalchemy.orm import joinedload, relationship
 
+
 from .amendement import Amendement
+from .article import Article
 from .base import Base, DBSession
+from .division import SubDiv
 from .journal import Journal
+
 
 CHAMBRES = {"an": "Assemblée nationale", "senat": "Sénat"}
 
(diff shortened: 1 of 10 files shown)


repondeur/zam_repondeur/models/lecture.py
```diff
@@ -236,3 +240,41 @@ def url_key(self) -> str:
...
+        if article is None:
+            article = Article.create(
+                lecture=self,
+                type=subdiv.type_,
+                num=subdiv.num,
+                mult=subdiv.mult,
+                pos=subdiv.pos,
+            )
+            created = True
+        return article, created
```
davidbgk: At that point, does `self.articles` contain the newly created article or does it need a DBSession flush/commit?

ronnix: Yes it does.

That's the beauty of relationships when working with objects (versus just by setting foreign keys).

If I do:
```python
article = Article(lecture=lecture, ...)
```
then this is true:
```python
assert article in lecture.articles
```

PR #341, opened 2019-02-28
JS update notifications for tables’ amendements

davidbgk:


davidbgk:
Review advice: [unified](https://github.com/betagouv/zam/pull/341/files?utf8=%E2%9C%93&diff=unified&w=1)

repondeur/zam_repondeur/views/tables.py
```diff
@@ -81,3 +84,14 @@ def post(self) -> Response:
...
         return HTTPFound(location=next_location)
+
+
+@view_config(context=TableResource, name="check", renderer="json")
+def table_check(context: TableResource, request: Request) -> dict:
+    table = context.model()
+    amendements_as_string = request.GET["current"]
+    updated = table.amendements_as_string
+    if amendements_as_string != updated:
+        return {"updated": updated}
```
ronnix: I understand that it's a way to reuse the existing notifications mechanism but I find it a bit confusing that `updated` is not the list of updated amendements, but the complete list, and the diff is computed client-side.

davidbgk: Indeed, my goal was to keep the order intact across the process for easier comparison of existing changes prior to go into the details but that might be a bad idea on the long term. 🤔

PR #352, opened 2019-03-19
Deal with abbreviated Gouv/Irr/Ret statuses

davidbgk:
Maybe a macro would be more pertinent?


ronnix:
Let's remove the dotted underlines

davidbgk:
Done!

PR #363, opened 2019-03-26
Make team names unique (with upgrade to SQLAlchemy 1.3.x)

ronnix:


davidbgk:
Is that defensive or it actually happened?

ronnix:
I don’t think it actually happened, and with the current code it’s probably not possible,  but when looking at the schema I thought it’s best to have an explicit constraint here.

ronnix:
(Similar to what we have on users.email)

PR #364, opened 2019-03-27
Avoid lonely identique amendements on index

davidbgk:
It currently happens when on of the identiques is retiré or irrecevable.


repondeur/tests/integration/test_amendements_identiques.py
```diff
@@ -24,3 +24,20 @@ def test_amendements_identiques(wsgi_server, driver, lecture_an, amendements_an)
...
+):
+    from zam_repondeur.models import DBSession
+
+    LECTURE_URL = f"{wsgi_server.application_url}lectures/{lecture_an.url_key}"
+    with transaction.manager:
+        amendements_an[0].id_identique = 42
+        amendements_an[1].id_identique = 42
+        amendements_an[1].sort = "retiré"
+        DBSession.add_all(amendements_an)
+        driver.get(f"{LECTURE_URL}/amendements")
```
davidbgk: There is something wrong here.

I guess we shouldn't have to call the driver within the transaction *but* `amendement.all_identiques` relies on `amendement.article` and it is detached to the session at that point and I wasn't able to figure out how to attach it.

😭

ronnix: Doing the get inside the context manager means the changes are not committed to the database when the browser does the request, so the test was passing for the wrong reason 😄

PR #372, opened 2019-03-31
Turn fetched dataclasses into regular models

davidbgk:
Currently missing:

* [ ] migrations
* [ ] decision to create all lecture then activate(== make visible) on user creation (might initiate that status thing with archived lectures)
* [ ] clean up the mess with Chambres and TypeTextes, turn these into models too?
* [ ] better use of new models (we used to iterate on all values)
* [ ] find a way to speed up tests with scoped fixtures

There are TODO to help, `pytest --stepwise` is your friend 😅

### DIFF
--- a/repondeur/tests/conftest.py
+++ b/repondeur/tests/conftest.py
@@ -1,4 +1,8 @@
+import json
 import os
+from datetime import datetime
+from pathlib import Path
+from unittest.mock import patch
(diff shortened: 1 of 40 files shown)


davidbgk:
Smaller step made here: https://github.com/betagouv/zam/pull/374

PR #373, opened 2019-04-01
Enable gzip compression in Nginx for non-HTML resources

ronnix:


ronnix:
@davidbgk CSS, JS, SVG...

ronnix:
@davidbgk Apparently the default is to gzip the `text/html` content-type only.

davidbgk:
Ah OK get it, I wasn't sure of the title given it compresses also the HTML

PR #382, opened 2019-04-08
User authentication (WIP)

ronnix:

PR #383, opened 2019-04-08
Add some style back for errors in login / welcome forms

ronnix:


ronnix:
Right now yes, but in the future (see #382) we may ask for a password and we'll need this

PR #390, opened 2019-04-08
Put irrecevables back to the index on fetch

davidbgk:


davidbgk:
Oh, I didn't think about the journal entry, good point!

PR #392, opened 2019-04-09
Auth refactorings

ronnix:
Extracted from #382 (WIP)


repondeur/tests/conftest.py
```diff
@@ -9,22 +10,30 @@
 from testapp import TestApp as BaseTestApp
 
 
+@contextmanager
+def auto_login(self, kwargs):
```
davidbgk: It looks weird to have `self` here, I would pass directly `self.user_login` to the manager.

ronnix: I’ll try something like that

ronnix: Or maybe we can move the function into the class and make it a method, then it's no longer weird to have `self`...

PR #407, opened 2019-05-06
Add fixtures closer to reality based on real data

davidbgk:


ronnix:
I think I'll just squash the commits, to keep only what's related to fixtures in the history.

davidbgk:
My intent was to use the Squash and merge button from Github, not enough?

ronnix:
Ah ok, did not think of that. I guess I’m used to the developer doing the squashing, not GitHub?

-- 
Ronan Amicel

davidbgk:
I didn't noticed you already did it, nevermind :)

ronnix:
Not sure GitHub can do the squash when there are merge commits in the branch, by the way.

PR #413, opened 2019-05-09
Prefetch URLs prior to set a lock on fetch

davidbgk:
Would hopefully allow to lock for a shorter amount of time on scheduled fetchs.


davidbgk:
I forgot to set it as a draft PR… WIP because we need to discuss the pertinence of the whole concept of dry run for that case @ronnix

davidbgk:
For the moment it's still pertinent but we spend a lot of time given the current 302 redirection on the AN side:

`http://www.assemblee-nationale.fr/15/xml/amendements/1408/AN/231.xml` == 302 => `http://www.assemblee-nationale.fr/dyn/15/amendements/1408/AN/231.xml` == 200

CacheControl is not caching 302 redirections (which is expected) but it leads to an extra request even if we have the end result cached.

Proposition: avoiding the redirection targeting directly the `/dyn/` URL and be reactive on changes 🙄

ronnix:
Did you try it? How much time does it take once the cache is filled if we bypass the 302s?

repondeur/zam_repondeur/tasks/fetch.py
```diff
@@ -38,24 +38,32 @@ def fetch_articles(lecture_pk: int) -> None:
 
 @huey.task(retries=3, retry_delay=RETRY_DELAY)
 def fetch_amendements(lecture_pk: int) -> None:
-    with transaction.manager, huey.lock_task(f"fetch-{lecture_pk}"):
```
ronnix: The thing we’re trying to make short here is the database transaction, not the Huey lock

davidbgk:
> Did you try it? How much time does it take once the cache is filled if we bypass the 302s?

The variability is quite high from that side of the Atlantic ocean :)

Without 302s: dry run is about 102 seconds and cached run is 22 seconds
With 302s: dry run is 125 seconds and cached run is 64 seconds

For a lecture with 313 amendements.

repondeur/zam_repondeur/tasks/fetch.py
davidbgk: 🤐

davidbgk:
To discuss: using https://github.com/ionrock/cachecontrol/blob/master/cachecontrol/heuristics.py#L100-L102

davidbgk:
After discussion with @ronnix, we think it's better to stay as it is instead of adding more complexity on the fetching part. At least for now.

PR #403, opened 2019-04-30
Remove modified_at and related notifications

ronnix:

PR #425, opened 2019-05-14
Moar events related to lectures (creation + imports)

davidbgk:
Capture d’écran, le 2019-05-14 à 16 35 00


repondeur/tests/test_lectures_add.py
```diff
@@ -279,6 +288,10 @@ def test_post_form_already_exists(
     assert resp.status_code == 200
     assert "Cette lecture existe déjà…" in resp.text
 
+    with transaction.manager:
```
ronnix: We probably don’t need to wrap this in a DB transaction, as there will be no changes to commit

repondeur/zam_repondeur/views/import_json.py
```diff
@@ -53,7 +53,7 @@ def import_backup(context: LectureResource, request: Request) -> Response:
         elif counter["articles"]:
             message = f"{counter['articles']} article(s) chargé(s) avec succès"
         request.session.flash(Message(cls="success", text=message))
-        lecture.modified_at = datetime.utcnow()
```
ronnix: We don’t need the modified_at any more?

davidbgk: AFAIK, [you removed it](https://github.com/betagouv/zam/pull/403) ;)

It was re-introduced here when I [restored the feature](https://github.com/betagouv/zam/pull/414) and given that if doesn't do any harm it stayed there silently since then.

repondeur/tests/test_lectures_add.py
davidbgk: Without the transaction I have a `sqlalchemy.orm.exc.DetachedInstanceError` when we access events.

repondeur/zam_repondeur/views/import_json.py
ronnix: Ok so let’s remove it a second time then!

repondeur/tests/test_lectures_add.py
ronnix: Even with just the DBSession.add?

davidbgk: Ah! You're right

repondeur/tests/test_upload_json.py
```diff
@@ -56,6 +56,19 @@ def test_upload_success_message(self, app, user_david):
         assert resp.status_code == 200
         assert "2 réponse(s) chargée(s) avec succès" in resp.text
 
+    def test_upload_success_event(self, app, user_david, lecture_an):
+        from zam_repondeur.models import DBSession
+
+        self._upload_backup(app, "backup.json", user_david).follow()
+
+        with transaction.manager:
```
ronnix: Same here?

PR #452, opened 2019-05-28
Remove Lecture.session

ronnix:


repondeur/zam_repondeur/models/texte.py
```diff
@@ -85,6 +85,12 @@ class Texte(Base):
         "date_depot",
     )
 
+    @property
+    def session_str(self) -> Optional[str]:
+        if self.session is None:
+            return None
```
davidbgk: If we don't return an empty string we will have `None` in URLs too?

ronnix: For AN, `Texte.session` is `None` but we have `Texte.legislature`. For Sénat, it's the opposite. So we should never have None in the Sénat fetch URLs.

davidbgk: 👌

repondeur/zam_repondeur/fetch/senat/derouleur.py
```diff
@@ -127,22 +127,23 @@ def _parse_derouleur_data(data_iter: Iterable[Any]) -> List[DiscussionDetails]:
 
 
 def derouleur_urls(lecture: Lecture) -> Iterator[str]:
+    assert lecture.texte.session_str is not None  # mypy hint
```
davidbgk: I'm surprised this line doesn't generate an error on the bandit side.

ronnix: Yup

ronnix: Rebase manually, solved the conflicts, and bandit was indeed unhappy with the asserts.

PR #459, opened 2019-05-29
Clean up requirements.txt files (case, ordering, missing transitive deps)

ronnix:


ronnix:
No, manually compared to pip freeze output to look for missing deps. But at some point I’d like to move to pip-tools.

PR #460, opened 2019-05-29
Fix potential security issues flagged by bandit (or mark them as safe)

ronnix:


ronnix:
I've either fixed the issues pointed by bandit or marked them as safe based on my understanding. Feel free to challenge it, and we can also review this in the next meeting of the Zam security department.

PR #485, opened 2019-06-04
Option 1: remove hypothesis-based test for batches

ronnix:
Unstable in CI, not sure why.

We may revisit using hypothesis later for this test or for others...


ronnix:
Going with option 2!

PR #533, opened 2019-06-13
Batches fixes, especially for amendements gouvernementaux

ronnix:


repondeur/zam_repondeur/views/lectures.py
```diff
@@ -327,25 +327,31 @@ def get(self) -> Any:
 
     @view_config(request_method="POST")
     def post(self) -> Response:
-        amendements = self.get_amendements_from(self.request.POST)
+        # Special case: unbatch (TODO: move to a separate route)
```
davidbgk: I think the checks must still be performed before unset-ing the batch?

davidbgk: OK forget about this 😅 💤

ronnix: Well, we could check it's on the user's table to allow removing it from the batch

PR #543, opened 2019-06-17
Refactor get_one_or_create

davidbgk:
* query `options` are now applied even on retry, was only on first attempt
* `kwargs` are only updated for creation, remains unchanged on subsequent get

I first tried to [mark the `IntegrityError` as retryable](https://docs.pylonsproject.org/projects/pyramid-retry/en/latest/#custom-retryable-errors) but it only applies to the local block and we want it to be applied to the whole get-or-create block. So I think there is no easy way of doing it with pyramid_retry but I'd love to be proven wrong :)


ronnix:
@davidbgk not sure what you mean by " it only applies to the local block" ?

davidbgk:
@ronnix when the `_create` fails, we want to retry the `_get_one` first, not the `_create`

ronnix:
@davidbgk my understanding is that `pyramid-retry` will retry the whole request, so it will go through the select query again before trying the create

ronnix:
👍

PR #553, opened 2019-06-25
Fix old migration

ronnix:


davidbgk:
Do we have to replay the migration everywhere?

ronnix:
No, it happened to me when replaying all migrations from scratch on a clean DB. Existing instances are already up to date.

PR #569, opened 2019-07-09
Deal with doslegs instead of unitary lectures

davidbgk:
- [x] switch from pk to slug for doslegs in URLs?
- [x] deal with ownership and doslegs
- [x] add a migration for ownership
- [x] create all lectures upon dosleg creation
- [x] remove unitary lecture creation
- [x] switch lecture delete to dossier delete
- [x] ~~fill the history (missing lectures for an existing dosleg)?~~ done by auto-update on doslegs
- [x] auto-update doslegs if new lectures are detected they should be created
- [x] create a dedicated team on dosleg activation
- [x] whitelist of administrators (us + selected SGG members)
- [x] UI to add/remove people to a dosleg/team
  - [x] invite first people by administrators by email invitation
  - [x] everybody from a dosleg can invite other people by email invitation

PR #580, opened 2019-07-16
Only enqueue huey tasks when the transaction is successful

ronnix:


repondeur/tests/conftest.py
```diff
@@ -112,6 +113,13 @@ def db():
     DBSession.remove()
 
 
+@pytest.fixture(autouse=True)
+def run_each_test_in_a_fresh_new_transaction():
```
davidbgk: Oooh, does it remove some “weirdness” in existing tests?

ronnix:
@davidbgk happy to explain the details, and yes this was prompted by issues with dossier creation in #569 ; did not think about the occasional weirdness in tests, but it may be related, and at best it cannot hurt as it gives us better isolation between tests

PR #591, opened 2019-07-23
Handle shared tables

davidbgk:


davidbgk:
Missing: event on shared tables CRUD.

PR #607, opened 2019-08-02
Reset Redis locks on initialization of data repository

davidbgk:


davidbgk:
@ronnix on Slack:

```
J’ai peur que dans le initialize si on restart la webapp ca pète les locks éventuellement mis par le worker 
Est-ce que faire ça comme une opération dans le déploiement résoudrait le problème de blocage ?
```

PR #676, opened 2019-09-05
Fix Redis key expiration timezone issue

ronnix:
The issue would only occur when the client and server would have different timezones (e.g. running tests locally against a redis server in a Docker container).

The redis-client library would compute the timestamp differently:

```python
>>> dt = datetime.now(tz=timezone.utc)
>>> int(mod_time.mktime(dt.timetuple()))
1567687995
>>> int((dt - datetime(1970, 1, 1, tzinfo=timezone.utc)).total_seconds())
1567695195
```
This would cause Redis to think the key expires in the past.

By computing the expiration timestamp ourselves, we can circumvent the problem and send Redis the right expiration time.


ronnix:
Merging this without pre-merge review

davidbgk:
👌 it took me a while to understand why the diff was solving the issue 😅

PR #706, opened 2019-09-20
Don't send exception to Rollbar if it's going to be retried

ronnix:


ronnix:
Squash ?

davidbgk:
Je vais Squash and merge via github

PR #712, opened 2019-09-23
Fetch credentials + X-Requested-With header

davidbgk:
First, without included credentials, old (60.7) versions of Firefox would not authenticate the request and thus record the current activity of the user.

Then, setting the `X-Requested-With` header will avoid recording an activity on fetch requests, which was giving a false sense of presence for tabs in background.


ronnix:
Fantastic!

PR #724, opened 2019-09-27
Do not count derouleur limit as an amendement

davidbgk:
Previously, the global number of amendements was incorrect with something as weird as `5 amendements (4 au total)`

Bug introduced in https://github.com/betagouv/zam/pull/723


davidbgk:
Post-review welcome!

PR #725, opened 2019-09-27
Only display the limit derouleur for displayables

davidbgk:
Avoid having multiple limits if the order is set but the amendements are abandoned.


davidbgk:
Post-review welcome!

PR #733, opened 2019-09-30
Bump tlfp targeting new TextParsingFailedException

davidbgk:
Cf. https://github.com/regardscitoyens/the-law-factory-parser/issues/126


repondeur/zam_repondeur/fetch/articles.py
```diff
@@ -70,7 +70,7 @@ def parse_first_working_url(urls: List[str]) -> List[dict]:
     for url in urls:
         try:
             articles: List[dict] = parse(url, include_annexes=True)
-        except Exception:
+        except TextParsingFailedException:
```
ronnix: Are we sure this is the only kind of error that can be raised by the `parse` function?

PR #745, opened 2019-10-03
Deal with senateurs and cache

davidbgk:


davidbgk:
There is a high probability the code would break if the matricule is absent from the data but it was already the case, not sure if we want to fix it (in that PR).

PR #759, opened 2019-10-08
Don't allow inviting non-ASCII email addresses

ronnix:
Fixes https://rollbar.com/zam/zam/items/360/

---

Capture du 2019-10-08 15-58-48


davidbgk:
I should explore a bit more more-itertools 😍

ronnix:
The more I look into it, the more I like it

PR #765, opened 2019-10-11
Security: addresses alerts from OWASP ZAP tool

davidbgk:


davidbgk:
More modifications based on https://ssl-config.mozilla.org/ + https://nginx.org/en/docs/http/ngx_http_ssl_module.html

davidbgk:
Still doubting about the pertinence of OCSP in our case, to be discussed.

PR #770, opened 2019-10-14
Remove Mission model

davidbgk:
Now denormalized as Amendement attributes

This way we don’t have to handle the complexity of labels evolutions across time.


davidbgk:
My take is that missions related data will fill up during the next refresh so I didn't migrate any data. Thoughts?

ronnix:
I think we should do a data migration to avoid the gap

PR #777, opened 2019-10-15
Dossiers Sénat

davidbgk:
TODO :
- [x] deduplicate Dossiers


ronnix:
Pushed some new commits. Still some work to do find the right Sénat dossier to scrape when the `senat_url` info is missing.

davidbgk:
` +2,000 −1,111 ` I just can't believe it. 😮

PR #800, opened 2019-10-23
Fix dossier locking in worker

ronnix:
Bad indentation after refactoring in 7f38244aab85d5769b09b6fca0ed0afd1783cdc2 prevented the body of the task from being protected by the task lock.

As a consequence, multiple worker threads could try acquiring an exclusive Postgres lock on the same row in the `lectures` table, causing blocked tasks.


davidbgk:
:+1:

PR #802, opened 2019-10-24
Configure graceful shutdown for the worker systemd service

ronnix:


davidbgk:
The autorebase bot doesn't look to be reliable anymore 🤷‍♂

PR #810, opened 2019-11-04
Make HTTP cache dir configurable in INI file

ronnix:
Most settings are in the INI file, but for some reason this one was using an environment variable.

(Also, it makes it easier to solve a pain point where in development I want to prevent Zam commands run in a container from creating files owned by root on the host.)


repondeur/zam_repondeur/services/fetch/http.py
```diff
@@ -1,16 +1,29 @@
...
+    Called automatically via config.include("zam_repondeur.services.fetch.http")
+    """
+    session = requests.session()
+    http_cache_dir = config.registry.settings["zam.http_cache_dir"]
+    cached_session = CacheControl(session, cache=FileCache(http_cache_dir))
+    config.registry.registerUtility(component=cached_session, provided=IHTTPSession)
 
-cached_session = CacheControl(session, cache=FileCache(HTTP_CACHE_DIR))
+
+def get_http_session(registry: Optional[Registry] = None) -> CacheControl:
```
davidbgk: Do we use that optional param somewhere?

repondeur/zam_repondeur/services/fetch/http.py
```diff
@@ -1,16 +1,29 @@
...
+    if registry is None:
+        registry = get_current_registry()
+    return registry.queryUtility(IHTTPSession)
```
davidbgk: What if we add `cached_session = get_http_session()` at the end of the file? I guess we don't want to instanciate it this early?

repondeur/zam_repondeur/services/fetch/http.py
ronnix: Not yet, but I feel less bad if I give the caller the opportunity of passing it as a param and avoiding a `get_current_registry()` call when not necessary...

(Similar to https://github.com/betagouv/zam/blob/master/repondeur/zam_repondeur/templating.py#L43-L45)

repondeur/zam_repondeur/services/fetch/http.py
ronnix: Exactly, we don't want to initialize it at import time (too early). We want to wait for the app to read its config and call the `includeme()` for initialization.

repondeur/zam_repondeur/services/fetch/http.py
davidbgk: Got it 👍

repondeur/zam_repondeur/services/fetch/http.py
davidbgk: OK

PR #825, opened 2019-11-06
Fix lecture sorting in PLF

ronnix:
Cannot compare int and None.


ronnix:
Should not happen really, but I ran into an error in development when I had created some bad lectures.

PR #842, opened 2019-11-13
Prefetch URLs prior to set a lock on fetch

davidbgk:
Inspired by https://github.com/betagouv/zam/pull/413


repondeur/development.ini
```diff
@@ -37,6 +37,8 @@ zam.auth_cookie_secure = false
 zam.refresh.amendements = 365
 zam.refresh.articles = 365
 
+# Minutes
+zam.http_cache_duration = 1
```
davidbgk: Not sure about the value we want for development.

repondeur/tests/fetch/mock_an.py
```diff
@@ -10,15 +10,22 @@ def setup_mock_responses(lecture, liste, amendements):
     from zam_repondeur.services.fetch.an.amendements import build_url
 
     with responses.RequestsMock(assert_all_requests_are_fired=False) as mock_resp:
+        # All responses are duplicated to handle the initial dry run
```
davidbgk: Another option would be to clear the cache after the dry run in tests like:

```python
from zam_repondeur.services.fetch.http import get_http_session
sess = get_http_session()
adapter = sess.get_adapter("http://")
adapter.cache.delete("http://www.assemblee-nationale.fr/eloi/15/amendements/0269/AN/liste.xml")
```

Both options look quite ugly to me :/

ronnix: Or can we just skip/disable the dry run in tests?

PR #845, opened 2019-11-14
Show exact time instead of "... ago" in journal entries

ronnix:
## Avant
Capture du 2019-11-14 17-42-28

## Après
Capture du 2019-11-14 17-41-58


davidbgk:
I would have done it in JS, for instance:

```diff
diff --git a/repondeur/zam_repondeur/static/js/zam.js b/repondeur/zam_repondeur/static/js/zam.js
index d66a819..46ec073 100644
--- a/repondeur/zam_repondeur/static/js/zam.js
+++ b/repondeur/zam_repondeur/static/js/zam.js
@@ -40,9 +40,16 @@ application.register(
       const { interval, epoch } = getDuration(timeAgoInSeconds)
       if (epoch === 'now') {
         return `à lʼinstant`
-      } else {
+      } else if (epoch === 'heure' || epoch === 'minute' || epoch === 'seconde') {
         const suffix = interval === 1 || epoch === 'mois' ? '' : 's'
         return `il y a ${interval} ${epoch}${suffix}`
+      } else {
+        return date.toLocaleTimeString('fr', {
+          month: 'short',
+          day: 'numeric',
+          hour: '2-digit',
+          minute: '2-digit'
+        })
       }
     }
   }
```

I'm OK with the python version though.

ronnix:
We also use the same JS to format the last modification time, and I think it works well as is for that, so maybe we could fork and have two different controllers, one for each context.

PR #872, opened 2019-11-22
Remove the useless tinysort JS dependency

davidbgk:
All this time we were loading it from index for nothing! 😱


ronnix:
Right, the feature was removed in 83e251fd96e8e531386633f68312a95b94f8ec5c but looks like we kept the dependency!

PR #888, opened 2019-11-28
Cache busting of tinymce w/ path update

davidbgk:
It appears that a proxy is caching URLs without considering the `x=` parameter we put to invalidate the cache


davidbgk:
Next step: https://docs.pylonsproject.org/projects/pyramid/en/latest/api/static.html

davidbgk:
The useful page is https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#cache-busting

PR #971, opened 2020-01-27
Fix error message when organe cannot be found in Open Data

ronnix:


repondeur/zam_repondeur/services/fetch/an/amendements.py
```diff
@@ -522,7 +522,7 @@ def get_organe_abrev(organe_uid: str) -> str:
 
     organe = repository.get_opendata_organe(organe_uid)
     if organe is None:
-        raise OrganeNotFound(organe)
+        raise OrganeNotFound(organe_uid)
```
davidbgk: 😅

PR #977, opened 2020-01-31
Remove Avis and Reponse columns from lecture index

davidbgk:


davidbgk:
This feature is essential to the users.

PR #978, opened 2020-02-03
Add script to add fake amendements to an existing lecture (for performance testing)

ronnix:


ronnix:
Sure, I wanted to push this first step quickly, but I'm sure we'll find ways to improve!

PR #984, opened 2020-02-04
Improve handling of URLs with very long query strings

ronnix:


ronnix:
Let's say a user already has the index opened in their browser, then we deploy the new version, then the user selects a few amendements and clicks "Transfer". In that case we get the old arg names.

PR #985, opened 2020-02-04
Fetch AN amendements in batches

ronnix:


repondeur/zam_repondeur/tasks/fetch.py
```diff
@@ -101,50 +104,75 @@ def fetch_amendements(lecture_pk: Optional[int]) -> bool:
...
-        # on some interactive operations).
+        nb_batches = 1
+        start_index = 0
 
-        # But during tests we want everything to run in a single transaction that
-        # we can roll back at the end.
-        if not huey.immediate:
-            transaction.commit()
-            DBSession.expire_all()
+        while True:
```
davidbgk: Given that we have a `total_timer` maybe we can add a timeout to that loop if it takes too much time?

repondeur/zam_repondeur/services/fetch/an/amendements.py
```diff
@@ -181,14 +200,16 @@ def _collect_amendements_other(
...
             except Exception:
-                logger.exception("Error while fetching amendement %r", numero)
-                errored.append(str(numero))
+                logger.exception("Error while fetching amendement %r", numero_prefixe)
+                errored.append(numero_prefixe)
                 continue
-            if numero > max_num_seen:
-                max_num_seen = numero
-        return actions, unchanged, errored
+        return actions, unchanged, errored, nb_404
```
davidbgk: At that point, we might want to return a structure? 🤔

repondeur/zam_repondeur/services/fetch/an/amendements.py
```diff
@@ -57,10 +61,26 @@ def __init__(self, organe: str) -> None:
...
+    def advance(self, offset: int) -> None:
+        self.lecture.set_fetch_progress(self.start_index + offset + 1, self.total)
+
+
 class AssembleeNationale(RemoteSource):
     def __init__(self, settings: Dict[str, Any], prefetching_enabled: bool = True):
         super().__init__(settings=settings, prefetching_enabled=prefetching_enabled)
+        self.batch_size = int(settings.get("zam.fetch.an.batch_size", BATCH_SIZE))
         self.max_404 = int(settings.get("zam.fetch.an.max_404", MAX_404))
+        if self.max_404 > self.batch_size:
```
davidbgk: I wonder if we can make that try at launch time?

repondeur/zam_repondeur/services/fetch/an/amendements.py
davidbgk: Mmmh re-taking a look at the real usage I'm not sure anymore…

repondeur/zam_repondeur/services/fetch/an/amendements.py
```diff
@@ -461,23 +480,56 @@ def build_url(
...
     """
-    Data extaction for Assemblée Nationale dérouleur
+    Data extraction for Assemblée Nationale dérouleur
     """
 
     def __init__(self, lecture: Lecture, content: dict):
         self.lecture = lecture
-        self.content = content
+        self.discussion_items = self._discussion_items(content)
+        self.position = self._position(self.discussion_items)
```
davidbgk: Why not `positionS`? If I understand correctly it stores a dict with all the positions?

ronnix: I refactored all this part

repondeur/zam_repondeur/services/fetch/an/amendements.py
```diff
@@ -57,10 +63,27 @@ def __init__(self, organe: str) -> None:
         self.organe = organe
 
 
+class ProgressBar:
```
davidbgk: I wonder if we should/could reuse the [ProgressBar from progressist](https://github.com/pyrates/progressist/blob/master/progressist/__init__.py#L47) (we already have that dependency) 🤔

ronnix: No idea

repondeur/zam_repondeur/tasks/fetch.py
ronnix: How long is too long?

davidbgk: 100000/250*120/60/60 => 13h20 if we consider a max loop at 2 minutes (120 seconds) 😰

PR #994, opened 2020-02-05
Tune Postgres configuration for increased performance

ronnix:
https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server


admin/tasks/system.py
```diff
@@ -43,15 +43,29 @@ def system(ctx):
 @task
 def setup_postgres(ctx):
     install_packages(ctx, "postgresql")
-    sudo_put(
-        ctx,
+    shared_buffers = total_memory(ctx) // 4  # 25% total RAM
```
davidbgk: Why 25%?

davidbgk: > If you have a system with 1GB or more of RAM, a reasonable starting value for shared_buffers is 1/4 of the memory in your system.

PR #996, opened 2020-02-06
Display index of amendements by article select

davidbgk:


davidbgk:
Note: I copy-pasted a bunch of code from `amendements.js` to `table.js` because I guess that implementations will diverge from now on given that we don't want to split the table by article(?).

davidbgk:
Switching to a regular PR given that the remaining perf issue is not related to that change. But, it will not show if that solution is pertinent because it's is as slow as before until we figure out a way to never load all amendements for a given lecture.

There is an access when we test if we have to display the Dossier de banc in the menu (we might want to display it whatever the case, relatively easy) and multiple accesses when we deal with similar amendements on index (harder, we can pair on that one).

PR #1002, opened 2020-02-07
Add french number formatting for amendement counts

ronnix:
100000 amendements => 100 000 amendements :ok_hand:


davidbgk:
Not super fond of adding a (quite big) dependency for that, what about `'{:,}'.format(1000000).replace(',', '\xa0')`?

ronnix:
I guess that would work!

PR #1012, opened 2020-02-11
Allow Redis background saves even without lots of free memory

ronnix:


ronnix:
Nah, your comment makes sense, we could avoid the duplication by writing the confing file then having procps reload it or something.

PR #1024, opened 2020-02-14
Set a config setting to hide exports options

davidbgk:


davidbgk:
This is not an issue, even with 40k amendements, I'll propose a PR to actually completely remove the feature.

PR #1026, opened 2020-02-14
Bump MAX_404 to 300

davidbgk:
sparta


davidbgk:
Deployed, let's keep a synced master

PR #1027, opened 2020-02-15
Bump request timeout from 5 min to 10

davidbgk:


davidbgk:
Deployed in production, let's keep `master` in sync

PR #1028, opened 2020-02-15
Switch to server-side search for amendements

davidbgk:
This is a first step, preparing the search from the Index. No UI on that side for now.

The size of the HTML produced is significantly smaller but the generation time is still quite high, I tried to tweak the DB query without any success :(


davidbgk:
Now the search form is available on all lecture-related pages.

PR #1087, opened 2020-03-13
Setup a workflow from conseil to lecture

davidbgk:


davidbgk:
The way we parse article content is still… basic but that might be already a good step :)

repondeur/tests/models/test_lecture.py
```diff
@@ -3,10 +3,67 @@
...
 import transaction
 
+from zam_repondeur.models.chambre import Chambre
+
 # We need data about dossiers, texts and groups
 pytestmark = pytest.mark.usefixtures("data_repository")
 
 
+@pytest.fixture
+def lecture_ccfp(db):
```
davidbgk: It's not clear to me if it is pertinent to reuse fixtures from https://github.com/betagouv/zam/pull/1087/files#diff-5c85a9c69aaac9918b965ba3589eecb7R23-R66

davidbgk:
Maybe we should merge that one and iterate on textes creation only 🤔

PR #1095, opened 2020-03-18
Ability to reorder textes for a conseil

davidbgk:


davidbgk:
Still unsure about keeping `Dossier.order` as optional, it might be better for Zam compatibility?

PR #1133, opened 2020-04-08
New design for navigation in the menu: breadcrumbs + back button

ronnix:
Capture du 2020-04-08 18-28-37


repondeur/zam_repondeur/views/amendements.py
```diff
@@ -39,6 +37,7 @@ def __init__(self, context: AmendementResource, request: Request) -> None:
         )
 
     @view_config(request_method="GET")
+    @view_config(request_method="GET", name="amendement_edit")  # backwards compat
```
davidbgk, line 40: I don't get why we need backwards compatibility here and below?

ronnix: Changed urls → open pages before deployment may become 404s after

ronnix:
Tomorrow is another day!

PR #1137, opened 2020-04-10
Visam drag&drop

davidbgk:


davidbgk:
https://htmldom.dev/drag-and-drop-element-in-a-list Never give up!

PR #1141, opened 2020-04-14
Add pluggable menu actions

ronnix:


ronnix:
Not Visam specific. I'll backport the system to Zam first, then add the Visam menu entry,

PR #1154, opened 2020-04-28
Visam: improve authorization (only allow certain actions for some user classes)

ronnix:
- Fix order of menu actions
- Use different accent colors for different classes of users
- Improve authorization: only allow certain actions for some user classes (still need some work)


ronnix:
@mergifyio rebase

